PR #4, opened 2015-03-08
Updt:GetGraphic Function. 

quillblue:
Users now can define their custom front color and background color.


codebude:
@quillblue I implemented your idea in this Commit https://github.com/codebude/QRCoder/commit/2fadc390b83c10f6653921d230d873405d5c5799 . It's a mixture of both ideas. So users can choose between no colors (then default b/w is used), colors with Color-class-type or colors in HTML hex formatted strings.

PR #29, opened 2016-04-12
Update Form1.cs

zizwiz:
Added button so I could save bitmaps


zizwiz:
Not sure all the items got in properly so if it has issues please contact me and I can e-mail files to you.

codebude:
Thanks for your input!

Issue #35, opened 2016-05-14
QRCoderConsole Demo Doesn't Take Inputs from Stdin


mishfit:
It would be great to be able to pipe input to QRCoderConsole.

I'd be glad to work on it.

codebude:
Implemented by #36 
Thanks for your help @mishfit

PR #36, opened 2016-05-15
QRCoderConsole: Read Standard Input to Generate a QRCode

mishfit:
Satisfies #35


codebude:
Thanks again for your effort!

PR #38, opened 2016-05-16
Bitcoin Payload Generator

mishfit:
Add support for [BIP 21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#Transfer amount/size) bitcoin payment address uri.

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -1,4 +1,9 @@
-﻿namespace QRCoder
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Globalization;
+
+namespace QRCoder
 {
     public static class PayloadGenerator
     {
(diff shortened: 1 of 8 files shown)


codebude:
The Bitcoin generator looks good, but just one question. Why did you use XUnit? I ask, because unit tests are also on my to do list for the library, but I had used MSTEST or maybe nUnit. Maybe you can change my decision with an good argument?

mishfit:
xunit is kind of the choice for future Microsoft testing (MSTest is going away and dotnet core is using xunit).

Other than that, when comparing xunit to nunit not having to decorate the class with a testing attribute is kind of nice (there are other benefits like async tests, which don't apply here).

QRCoder/PayloadGenerator.cs
```diff
@@ -1,4 +1,9 @@
-﻿namespace QRCoder
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Globalization;
```
mishfit: I don't need System.Globalization after all

mishfit:
let me remove an extra using statement that I didn't need to add after all

mishfit:
let me know if you want the commits squashed

codebude:
Ok, sounds nice. I'll have a look into xUnit the next days.

codebude:
Hi @mishfit ,

I just ran your unit tests and two of them failed. For example:

Expected output:
`bitcoin:175tWpb8K1S7NmH4Zx6rewF9WQrcZv245W?label=Some%20Label%20to%20Encode&message=Some%20Message%20to%20Encode&amount=.123`
Actual output:
`bitcoin:175tWpb8K1S7NmH4Zx6rewF9WQrcZv245W?label=Some%20Label%20to%20Encode&message=Some%20Message%20to%20Encode&amount=,123`

The difference lies in the numeric notation. In germany decimals are seperated by , and not by .

So, what does the Bitcoin uri schema say? If there must be a . in every case, we maybe should change
`new Tuple<string, string>(nameof(amount), amount.HasValue ? amount.Value.ToString("#.########") : null)`
to
`new Tuple<string, string>(nameof(amount), amount.HasValue ? amount.Value.ToString("#.########").Replace(",",".") : null)`

What do you think?

mishfit:
Interesting.,.i'll have to look up what internationalized clients are doing

codebude:
Seems like it must be a dot (.) in every case.

> If an amount is provided, it MUST be specified in decimal BTC. All amounts MUST contain no commas and use a period (.) as the separating character to separate whole numbers and decimal fractions. I.e. amount=50.00 or amount=50 is treated as 50 BTC, and amount=50,000.00 is invalid.
*(Source: https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#General_Format)*

mishfit:
Well that pretty much clears that up...I'll have to force a format provider that doesn't use comma

codebude:
`Just replace the output with `.Replace(",",".")` or add cultural info to the ToString() function like `amount.Value.ToString("#.########", new CultureInfo("en-US"))`.

mishfit:
Yep, I like CultureInfo for that

PR #41, opened 2016-06-13
fix parameter

michaelchong2005:
It need pass the correct value of `drawQuietZones` to function `GetGraphic`.


codebude:
Thanks for your effort! :+1:

PR #52, opened 2016-09-21
lowered .net target version to 3.5cp

krisrok:
add minor changes to make the main project 3.5 compliant (so it can be used in e.g. unity3d)


codebude:
Thanks for your help!

PR #55, opened 2016-11-17
Refactored GetEncodingFromPlaintext.

dlemstra:
This is a minor refactor/optimization of the GetEncodingFromPlaintext method.


codebude:
Hey @dlemstra ,

thanks for you pull request. Just two questions before I merge your request:

1) Why did you remove the line `this.alphanumEncTable = null;` from the _Dispose()_-method? I don't see why it shouldn't stay there.
2) Why did you keep the _StringAll_-method? (Because with your changes it isn't needed any longer.)

Kind regards,
Raffael

dlemstra:
The StringAll method was left there by accident. Just did a new forced push to remove it.

alphanumEncTable can no longer be set to null because I made it static and readonly. It can be shared by all instances and that is why I made it static.

codebude:
Sounds legit. I'll merge the request. Thanks again.

PR #61, opened 2017-03-01
BML is not BMP

Hydroque:
This is to promote correctness.


codebude:
Could you please explain your change? You changed the third bit of the bitmap file header from 0x4C to 0x50. This byte is the first part of the "size"-block in bitmap file header. So how does the belong to your title: "bml is not bmp"?

Hydroque:
Nevermind c:

codebude:
So this Pull request is obsolete?

Hydroque:
Yep. My bad.

PR #87, opened 2017-09-19
Fixed rendering of SVG using SvgQRCode

kassovic-f:


codebude:
Hi @kassovic-f ,

can you explain your changes? I tried to compare the result of your code and mine code.
For "drawQuietZones=true" the result is the same. For "drawQuietZones=false" I see a difference.

Unfortunately your result can't be rendered. I tested the SVG results with:
http://www.freecodeformat.com/svg-editor.php
There your version/code doesn't render. 

So I would be happy if you could tell me, what you've changed and why you did this. Because right now, I can't see the benefit.

kassovic-f:
Original code problems:
image

Original rendering issues:
drawQuietZones = true:
image

drawQuietZones = false:
image


hope this helps.

Yep, u r right, I've just added CultureInfo.InvariantCulture. Sorry for that.

codebude:
Thanks for your valuable input! There were really some errors. When I implemented it, I used integers, because when using fractions/decimals The QR codes were rendered with tiny whitespaces between the blocks. But today I learned, that this can be avoided by adding `shape-rendering="crispEdges"` to deactivate antialiasing, which was the source for the whitespaces...

Since I wanted to keep the original parameter names, I decided to implement your input/tips myself instead of merging. See here: https://github.com/codebude/QRCoder/blob/master/QRCoder/SvgQRCode.cs 
Hope you're not said about this, but the parameter names are 1) consistent for all QrCode rendering types, so that I like to keep it simple and 2) partly official names. (QuietZone is a term from the QR code ISO norm. Yes, border seems to be more precise, but the official name is "quiet zone".)

kassovic-f:
Hi, no problem. I didn't know about this ISO terminology. Original request for the change of quiet zones switch also called it "border", so I thought that it is better as the change has no breaking change issues. :-)
One note: instead of
`input.ToString().Replace(",", ".");`
u should use `input.ToString(System.Globalization.CultureInfo.InvariantCulture)` as I mentioned before. There could potentially be other separators than "," or "."...

codebude:
Another good input from your side. Thanks. https://github.com/codebude/QRCoder/commit/544c3a1f2878d0f7f0728915740b38badd089f44

PR #109, opened 2018-03-07
feat: Connect from an oleobject using COM connection (Powerbuilder, VB)

### LINKED ISSUE #108: Connect from an oleobject using COM connection (VB 6, Powerbuilder...)
## Type of issue
[ ] Bug  
[X] Request for new feature/improvement  

## Expected Behavior

Connect from an oleobject using COM connection (VB 6, Powerbuilder...)

## Current Behavior

To connect from an oleobject from VB 6 or Powerbuilder, we need a constructor without parameters since it is not possible to parameterize it.

In addition, COM connections are not compatible with generic collections, throws an exception "function name not found".

This can be solved by eliminating the generic collections, since it would not affect the current functionality, and creating a constructor without parameters. Subsequently, the QRCodeData object that is currently assigned in the constructor with parameters, would be added manually with SetQRCodeData function

I will create a Pull Request with this new feature

## Your Environment

Version used: 1.3.3
Environment: Powerbuilder 6.5 & Powerbuilder 10.5

abelrodriguez:

* [X] Feature: Connect from an oleobject using COM connection

To connect from an oleobject from VB 6 or Powerbuilder, we need a
constructor without parameters since it is not possible to parameterize
it.

In addition, COM connections are not compatible with generic collections,
throws an exception "function name not found".

This can be solved by eliminating the generic collections, since it would
not affect the current functionality, and creating a constructor without
parameters. Subsequently, the QRCodeData object that is currently assigned
in the constructor with parameters, would be added manually with
SetQRCodeData function

**Closing issues**

closes #108


codebude:
Hi @abelrodriguez ,

thanks for this PR. It's just incredible - I discussed about implementing an COM Interop interface for QRCoder round about a hour ago with a colleague - and then came your PR...

Just let me ask one, two questions before merging:
- Why did you remove the Dispose methods? ( Which where added in this PR: https://github.com/codebude/QRCoder/pull/18 )
- Do you plan to deliver another PR for the PayloadGenerator class? (Otherwise I'll add it to my "to do"-list. ;-) )

abelrodriguez:
Hi @codebude,

I removed the Dispose methods because thought it would be more efficient integrate it in AbstractQRCode but I think I had an error on interface implementation/declaration.

About second question... =) At first sight, I think it could be complex. Maybe we have to do a code restructuring to extract each class declared within static class PayloadGenerator. So, for now, I don´t plan to deliver another PR unless you need it. In that case, I promise to deliver those changes.

Thank you very much for considering my PR.

PR #111, opened 2018-04-09
added SlovenianUpnQr payload generator, manual QR version and ECI mode

ahotko:
This PR implements the following **features**:

* [ ] Added ECI Mode Support (partial)
* [ ] Added possibility of Manual QR Code Version
* [ ] Added Slovenian UPN-QR Code standard Payload Generator (SlovenianUpnQr)

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -584,7 +584,7 @@ namespace QRCoder
                 return $"bitcoin:{address}{query}";
             }
         }
-        
+
         public class SwissQrCode
         {
             //Keep in mind, that the ECC level has to be set to "M" when generating a SwissQrCode!
@@ -615,7 +615,7 @@ namespace QRCoder
             public SwissQrCode(Iban iban, Currency currency, Contact creditor, Reference reference, Contact debitor = null, decimal? amount = null, DateTime? requestedDateOfPayment = null, Contact ultimateCreditor = null, string alternativeProcedure1 = null, string alternativeProcedure2 = null)
             {
                 this.iban = iban;
-                
+
                 this.creditor = creditor;
                 this.ultimateCreditor = ultimateCreditor;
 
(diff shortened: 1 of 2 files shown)


codebude:
Hi @ahotko ,

thanks for your effort. Please give me a day or two to review the code and think about it. (I want to think about all possible side effects and conflicts. But from a first view, I think I'll merge it. 👍 )

Another question: Would you mind, to deliver a text for the wiki, which explains the SlovenianQRCode? Similiar to the SwissQrCode part over here: https://github.com/codebude/QRCoder/wiki/Advanced-usage---Payload-generators

ahotko:
Hei,

I tried to change the code as little as possible so the functionality SHOULD be compatible or the same.

SlovenianUpnQr generates string according to Slovenian UPN QR standard (UPN is acronym for "Univerzalni placilni nalog" or in english "Universal payment order"). Requirements are in https://www.upn-qr.si/uploads/files/NavodilaZaProgramerjeUPNQR.pdf document, but it is in Slovene. It has some special requirements though, which have to be met to get approval for printing QR code in UPN forms. Theses requiremenst are:
- QR Version must be 15
- ECC must be M
- Byte mode
- ECI set to 4 (ISO8859-2)
- and of course specially formatted string

Attached if text for Wiki.

[SlovenianUpnQr-md.txt](https://github.com/codebude/QRCoder/files/1894645/SlovenianUpnQr-md.txt)

codebude:
Hi @ahotko ,

at first thanks for providing the documentation.md!  As promised I had a look over the code. Before merging I have  one or two questions:

- You added a method *PlainTextToBinaryECI*, but I can't see that it is used anywhere? Did I miss the part, where it is used or can it be removed from the source again?
- In the *PlainTextToBinary*-method you added a case-switch for EncodingMode.ECI, but added no return value (see here: https://github.com/codebude/QRCoder/pull/111/files?diff=split#diff-9e6c145bebc51d16883ade6f68b714dfR908 ). How does this work? Wouldn't this empty the payload?

Besides this two questions I'm fine with your changes. Could you help to erase my question marks?

ahotko:
Hei,

- _PlainTextToBinaryECI_  was ment to be for ECI Encoding, but as ECI is only for character encoding, mode doesn't matter (even better if it's byte mode), so I forgot to delete the method

- in case of empty _PlainTextToBinary_ case statement: _PlainTextToBinaryECI_ shoul be used here, but I have removed it, as it is not used; case can be empty as it defaults to default (as it did before) - you can remove it or leave it, does not matter as EncodingMode.ECI is never assigned (maybe future use?)

Kind regards,
Ales

codebude:
Ok,  the evrything is clear. Thanks for your support.  :+1:

codebude:
Hi @ahotko I just released [version 1.3.3 on Nuget](https://www.nuget.org/packages/QRCoder/). So your payload generator is part of the official Nuget package now.

ahotko:
Hei @codebude - thanks for info and merge! Have a great day!

PR #125, opened 2018-05-24
CreateAntilogTable doing unnecessary Pow calculations

SteveGTR:
This change doesn't modify the galoisField values as the previous version worked just fine in regard to the data generated. It was just doing an unnecessary Pow() calculation that was meaningless after the value of "i" was greater than 7.

* [ ] Bug 1
* [ ] Bug 2
* [ ] Feature 1
* [ ] Feature 2
* [ ] Breaking changes

**Test plan**

**Closing issues**

Fixes #


codebude:
Hi @SteveGTR ,

that's a valid point. I'll merge your changes.

PR #132, opened 2018-06-18
add an easy-to-call one-hit wrappers to generate PNG and BMP image byte array

sbrickey:
* Feature : single static methods - no instances, no disposables, just call and go.

### DIFF
--- a/QRCoder/BitmapByteQRCode.cs
+++ b/QRCoder/BitmapByteQRCode.cs
@@ -99,7 +99,7 @@ namespace QRCoder
     }
     
   
-    public sealed class BitmapByteQrCodeUtility
+    public static class BitmapByteQrCodeUtility
     {
         public static byte[] GetBmpQrCode(string txt, QRCodeGenerator.ECCLevel eccLevel, int size)
         {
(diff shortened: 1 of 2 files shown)


AdamWillden:
Oh, although I would appreciate if there were overloads for passing in a payload rather than just the plaintext

codebude:
Hi @sbrickey ,

I understand your idea/concept, but I don't like to merge it now for sake of completeness. I think if we add this feature to the QRCoder it should be consistent over all QR renderer classes and not only for some (like it is now). I fear that users could get confused if only some renderer support this "shortcuts". 

Would you mind added this shortcuts for the other rendering classes, too? If so, I would be happy to merge it.

Ruffio:
@sbrickey would you like to implement the suggestions by @codebude 
I don't think this PR will be merged without these additions ;-)

codebude:
Finally merged. Thanks for your support. Feel free to add some parts on the https://github.com/codebude/QRCoder/wiki about the new classes @eulercamposbarros , @Ruffio , @sbrickey .

PR #147, opened 2018-10-27
fixed issue #145 with OutOfRange Exception generating Xaml QR Code images

### LINKED ISSUE #145: XamlQRCode crash
## Type of issue
[X] Bug  

## Expected Behavior
XamlQRCode should not crash ...

XamlQRCode loops on the size adding doubles together hoping to get to the width/height. Adding doubles don't always round to the value you would normally expect. You should loop on the QRCode size and add pixel for placement instead.
```
double x = 0;
double y;
for (int xi = offsetModules; xi < QrCodeData.ModuleMatrix.Count - offsetModules; ++xi)
{
    y = 0;
    for (int yi = offsetModules; yi < QrCodeData.ModuleMatrix.Count - offsetModules; ++yi)
    {
        if (this.QrCodeData.ModuleMatrix[yi + offsetModules][xi + offsetModules])
        {
            group.Children.Add(new RectangleGeometry(new Rect(x, y, unitsPerModule, unitsPerModule)));
        }
        y += unitsPerModule;
    }

    x += unitsPerModule;
}
```

```
var generator = new QRCodeGenerator();
var data = generator.CreateQrCode("Something that ends up with a size 41 QRCode", QRCodeGenerator.ECCLevel.Q);
var xaml = new XamlQRCode(data);
Code = xaml.GetGraphic(new Size(200, 200), Brushes.Black, Brushes.White);
```

## Your Environment
Version Used: QRCoder 1.3.3 from NuGet

DevonJSmith:
**Summary**
Refactors the GetGraphic loop for Xaml QR Code to avoid a System.OutOfRange Exception, especially when dealing with `unitsPerModule` values that are doubles with repeating decimal places.

An example of this found in the original bug issue was that the size limit was `200`, but the unlimited precision of the double `yi` allowed it to reach a value of `199.99999992` (rough estimate) and so the loop did not terminate as expected.

This refactor instead loops on the data of the QR code (always an `int`) while adding to the pixel values, a reverse of the original structure. This avoids the System.OutOfRange exception when accessing the QR code data.

This PR fixes the following **bugs**:

* [x]  #145

Fixes OutOfRangeException encountered in #145

**Test plan**

Tested with repeatable bug code found in #145.
 Additionally, compare the image results from the QRCode library (base case) and the XamlQRCode library (using fix contained in this PR). Both were generated with the original bugged string "Something that ends up with a size 41 QRCode" : [imgur link](https://imgur.com/a/YUyZAuG)

**Closing issues**

`closes #145`

### DIFF
--- a/QRCoder/XamlQRCode.cs
+++ b/QRCoder/XamlQRCode.cs
@@ -48,19 +48,19 @@ namespace QRCoder
             drawing.Children.Add(new GeometryDrawing(lightBrush, null, new RectangleGeometry(new Rect(new Point(0, 0), new Size(qrSize, qrSize)))));
 
             var group = new GeometryGroup();
-            int xi = 0, yi = 0;
-            for (var x = 0d; x < qrSize; x = x + unitsPerModule)
+            double x = 0d, y = 0d;
+            for (int xi = offsetModules; xi < drawableModulesCount; xi++)
             {
-                yi = 0;
-                for (var y = 0d; y < qrSize; y = y + unitsPerModule)
+                y = 0d;
+                for (int yi = offsetModules; yi < drawableModulesCount; yi++)
                 {
-                    if (this.QrCodeData.ModuleMatrix[yi + offsetModules][xi + offsetModules])
+                    if (this.QrCodeData.ModuleMatrix[yi][xi])
                     {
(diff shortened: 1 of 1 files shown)


codebude:
Thanks for your input and help!

PR #148, opened 2018-10-27
Issue 139 rgb color generation

### LINKED ISSUE #139: BitmapByteQRCode.HexColorToByteArray swaps R and B in RGB colors
## Type of issue
[x] Bug  

## Expected Behavior
`BitmapByteQRCode.GetGraphic(20, "#ff0000", "#ffffff")` should generate a red graphic, just like `BitmapByteQRCode.GetGraphic(20, new byte[] { 0xff, 0x00, 0x00 }, new byte[] { 0xff, 0xff, 0xff })`.

## Current Behavior
The graphic is generated as blue (`#0000ff`).

Change https://github.com/codebude/QRCoder/blob/master/QRCoder/BitmapByteQRCode.cs line 85 to set `byteColor[i]` instead of `byteColor[2-i]`

```

        private IActionResult GetPlainQRCode(string url)
        {
            PayloadGenerator.Url generator = new PayloadGenerator.Url(url);
            string payload = generator.ToString();

            using (var qrGenerator = new QRCodeGenerator())
            {
                using (var qrCodeData = qrGenerator.CreateQrCode(payload, QRCodeGenerator.ECCLevel.Q))
                {
                    using (var qrCode = new BitmapByteQRCode(qrCodeData))
                    {
                        var qrCodeAsByteArray = qrCode.GetGraphic(20, "#FF0000", "#FFFFFF");
                        return File(qrCodeAsByteArray, "image/bmp");
                    }
                }
            }
        }
```

## Your Environment

* Version used:  1.3.3 (it looks like 1.3.4 was set in April but has not been released on nuget, is that right?)
* Compiled from source or NuGet package?:  Nuget
* Environment (.NET 3.5, .NET 4.X, .NETCore, ...):  .NET Core 2.1

DevonJSmith:
**Summary**

Small change to `HexColorToByteArray` method in `BitMapQRCode.cs` which fixes parsing of RGB values from hex strings.

Previously this was done backwards, so the value would be read as BGR instead of RGB.

In #139, this manifested noticeably where `#FF0000` would be output to an image as `#0000FF`

This PR fixes the following **bugs**:

* [x] #139 (BitmapByteQRCode.HexColorToByteArray swaps R and B in RGB colors)

**Test plan**

Tested with case identified in #139, generating an image with color: `#FF0000` has expected color on output: ([imgur link](https://imgur.com/a/Twx0Iom))

**Closing issues**

`closes #139`

### DIFF
--- a/QRCoder/BitmapByteQRCode.cs
+++ b/QRCoder/BitmapByteQRCode.cs
@@ -82,7 +82,7 @@ namespace QRCoder
                 colorString = colorString.Substring(1);
             byte[] byteColor = new byte[colorString.Length / 2];
             for (int i = 0; i < byteColor.Length; i++)
-                byteColor[2-i] = byte.Parse(colorString.Substring(i * 2, 2), System.Globalization.NumberStyles.HexNumber, System.Globalization.CultureInfo.InvariantCulture);            
+                byteColor[i] = byte.Parse(colorString.Substring(i * 2, 2), System.Globalization.NumberStyles.HexNumber, System.Globalization.CultureInfo.InvariantCulture);            
             return byteColor;
         }
 
--- a/QRCoderProject.NET40.sln
+++ b/QRCoderProject.NET40.sln
@@ -1,9 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+# Visual Studio 15
+VisualStudioVersion = 15.0.28010.2048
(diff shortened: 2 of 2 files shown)


codebude:
Thanks for fixing this. By the way the order BGR is correct for use in Bitmaps. Windows bitmaps have the color bytes in this order, thus the old code was actually ok. But what I had overseen is that I reversed the order also here: https://github.com/codebude/QRCoder/blob/master/QRCoder/BitmapByteQRCode.cs#L33 
So it was flipped back from the needed BGR to RGB...

PR #159, opened 2018-12-02
Easy to call wrappers to generate QRCode

eulercamposbarros:
**Summary**

Create static methods to generate QRCodes.

* [X] Feature: Static Helper

I saw a pending PR from months ago and thought about helping.

**Test plan**

I don't wrote any tests because to test static methods is not an easy task. But I might give a try if necessary.

**Closing issues**

Fix #132


codebude:
Merged. Thanks for your support @eulercamposbarros !

PR #161, opened 2019-01-26
Add contact/phone number to WhatsApp Payload

taqabubaker:
This PR gives creates a new overload to the WhatsApp payload by giving the user the ability to add contact/phone number to WhatsApp Payload. This is can be achieved since WhatsApp is allowing passing the contact/phone number in the following format
`$"whatsapp://send/?phone={number}&text={message}"`

This PR fixes/implements the following **features**:

* [X] Add contact/phonenumber to WhatsApp Payload

Just adding gives the user a new feature to send the WhatsApp message directly to the desired contact.

**Test plan**

UnitTests:

- [ whatsapp_generator_can_generate_payload_simple] still working.

- [ whatsapp_should_add_unused_params] new unit test method to test if the phone number is missing. 

**Closing issues**

Fixes #

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -326,7 +326,18 @@ namespace QRCoder
 
         public class WhatsAppMessage : Payload
         {
-            private readonly string message;
+            private readonly string number, message;
+
+            /// <summary>
+            /// Let's you compose a WhatApp message and send it the receiver number.
+            /// </summary>
+            /// <param name="number">Receiver phone number</param>
+            /// <param name="message">The message</param>
+            public WhatsAppMessage(string number, string message)
+            {
+                this.number = number;
+                this.message = message;
+            }
 
             /// <summary>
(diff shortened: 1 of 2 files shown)


codebude:
Thanks for your work! :+1:

taqabubaker:
@codebude 
I believe we need to change the wiki page [https://github.com/codebude/QRCoder/wiki/Advanced-usage---Payload-generators#319-whatsappmessage](url)

Following should be changed. What do you think ?

> Good to know
Unfortunately it is impossible to handover a contact/phonenumber, because WhatsApp only accepts its internal contact ids, which can't be known at time of QR code creation.

codebude:
@taqabubaker You're right. In addition the table with the overloads has to be expanded. Or should I say had... Have a look at the wiki - I already updated the article. ;-)

taqabubaker:
@codebude don't forget to publish a new version of the nuget package. :wink:

codebude:
@taqabubaker I will push a new version during the next week. I like to check the other issues in front of. Maybe I can fix one or more things. (I don't want to push multiple packages in a week, but prefer to put as much as possible into the releases.)

PR #165, opened 2019-02-15
Add support for binary data

oheim:
**Summary**

Adds a new method `public QRCodeData CreateQrCode(byte[] binaryData, ECCLevel eccLevel)` to encode binary data.

I need to encode binary data containing null-bytes. This is impossible with the string-based `CreateQrCode` methods, since the data would be truncated.

### DIFF
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -72,6 +72,27 @@ namespace QRCoder
 
             bitString += codedText;
 
+            return CreateQrCode(bitString, eccLevel, version);
+        }
+
+        public QRCodeData CreateQrCode(byte[] binaryData, ECCLevel eccLevel)
+        {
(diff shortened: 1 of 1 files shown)


codebude:
Hi @oheim can you explain what you mean by "This is impossible with the string-based CreateQrCode methods, since the data would be truncated."? Do you mean a string can't hold that much data like a byte array? If so, I think a byte array isn't the solution because a QR code by definition only can hold up ~4KB of data which can be stored without problems within a string.

oheim:
For example, consider the following data
````
var binaryData = new byte[] { 0x00, 0x01, 0x02, 0x03 };
````

It is impossible to pass this data as a string parameter, because strings are null-terminated. If the data to be encoded contains the byte `0x00` I cannot use the string based methods and have to use the new method.

codebude:
Ok, now I got it. Use case seems legit, so I'll merge the changes.

jnm2:
I really appreciate this! It's just what I needed today.

PR #169, opened 2019-03-09
The website field is not added to the vcard if it was of type other than MeCard

kashif-khan:
Root cause: Null was checked on phone field to add the website field to the vcard

* [ ] #170 

The website field was never added to the vcard if it was of type other than MeCard and if the phone field was null.

**Test plan**

Existing test cases should pass.

**Closing issues**

Fixes #
Changed the null checking to website field for adding the website data to the vcard.


codebude:
Thanks for fixing this bug! :+1:

PR #182, opened 2019-05-13
Feature/fix memory leaks

marcelhintermann:
This is just a small fix against some memory leaks.
Some System.Drawing Objects were not disposed.


codebude:
Thanks for your help!

marcelhintermann:
Thanks for merging!
could you please release a nuget package with the latest changes?

codebude:
Hi @marcelhintermann - I'll add one, two more changes this evening and then release a new version of the package.

PR #187, opened 2019-06-08
Add XamlQRCode.GetUnitsPerModule() method

rozmansi:
**Summary**

This is a helper method allowing client to obtain an exact measure of module size without a need to repeat the calculation by itself.

When the client wants to draw a border around QR code or some other adornments which need to use the same size and/or spacing as the module size this commit simplifies its calculation.

* [ ] Introduces XamlQRCode.GetUnitsPerModule() helper method


codebude:
Thanks for your support!

PR #191, opened 2019-07-11
Feature/ Added a flag to prevent/force Data trimming while creating QRCode while creating QRCode with requested version

riteshbxr:
**Summary**

Added a flag to prevent/force Data trimming while creating QRCode with requested version

* [x] When a Requested Version is put in Qr code Trims Data for the version, It should Auto Upgrade or User Should have an option to choose to trim or AutoUpgrade Version to keep Data in QR Code.

>What problem does it Solve
There is no option to set minimum version and so user may have to calculate Version himself. To prevent this, an option to send requested version and a flag to override/Enforce the Requested version to prevent data from getting trimmed off has been added.

**Test plan**
No functional changes. All existing tests pass. All old functioning have been ensured during the changes. Issue has been handled by adding another optional Boolean flag

### DIFF
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1,9 +1,8 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using System.Collections;
-using System.Reflection;
 
 namespace QRCoder
 {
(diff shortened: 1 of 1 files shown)


csturm83:
Is there actually a scenario where truncating the data is the expected/preferred behavior? I'm struggling to think of one.

Maybe instead of adding a flag, we should only respect the requested version if it is valid given the length of the data? 

I understand that will diverge from the existing behavior, but maybe it's more expected?

riteshbxr:
Exactly, by I did not want to change the behavior... I can set the default to allow version upgrade... If you allow me...

riteshbxr:
So, should I send pull request with changes?

codebude:
I though a while about these changes now and I think I won't merge them, because they don't make that much sense from my perspective.

By adding this additional parameter, one could avoid that a "DataTooLongException" is thrown, when the manually set version is to low for the data handed over. But does this make sense?

In most cases users won't set the version but let the library calculate the optimal version (smallest possible version which can handle the given payload/data). The only reason one sets a fixed version, is when one expects all QR codes to be the same block/module size. (If the module size doesn't matter, it makes no sense to set a fixed version and thus waste space/error tolerance.)

So if one sets a fixed version because his application is in need of every QR code to have the same amount of modules/blocks, it doesn't make any sense when the library could create higher version QR codes. This would break the concept of a fixed version. Throwing an exception telling the user that he put in a too big payload is (from my understanding) the only logical way.

If one nevertheless (and as I told - I don't see practical use cases) want's your behaviour he could just write a workaround like that:

```
var qrGenerator = new QRCodeGenerator();
QRCodeData qrCodeData;
try 
{
   qrCodeData = qrGenerator.CreateQrCode("My payload to be encoded.", requestedVersion: 1);
} 
catch (QRCoder.Exceptions.DataTooLongException dataEx)
{
   qrCodeData = qrGenerator.CreateQrCode("My payload to be encoded.");
}
```
So I think for these rare cases, we shouldn't add an additional parameter but just ask the users to use the workaround shown above. Also changing the behaviour by default (to use a bigger version if data is too long) is something I won't implement, because it would break the concept of a *fixed* version parameter.

PR #192, opened 2019-07-15
Update System.Drawing.Common dependency version in nuspec

### LINKED ISSUE #190: Nuspec out of date? (System.Drawing.Common)
May be experiencing a bug, or maybe not. Either way I've noticed a version mismatch you may want to get in sync:

https://github.com/codebude/QRCoder/search?q=system.drawing.common&unscoped_q=system.drawing.common

image

The issue I'm being presented with is this:

`Assembly 'QRCoder' with identity 'QRCoder, Version=1.3.6.0, Culture=neutral, PublicKeyToken=null' uses 'System.Drawing.Common, Version=4.0.0.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' which has a higher version than referenced assembly 'System.Drawing.Common' with identity 'System.Drawing.Common, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'`

I'm not using system drawing common anywhere in my project (a .Net Core 2.2 web project).

thomaslevesque:
To match version specified in csproj.
Also remove netstandard1.6, since the library doesn't actually target it (I assume copying from netstandard2.0 to lib/netstandard1.6 was a mistake, unless it's something clever that I totally misunderstood)

Fixes #190

### DIFF
--- a/QRCoder.nuspec
+++ b/QRCoder.nuspec
@@ -15,11 +15,8 @@
 		<copyright>Copyright &#x00A9; www.code-bude.net / Raffael Herrmann.  All rights reserved.</copyright>
 		<tags>c# csharp qr qrcoder qrcode qr-generator qr-code-generator</tags>
 		<dependencies>			
-			<group targetFramework="netstandard1.6">
-				<dependency id="System.Drawing.Common" version="4.5.0-preview1-25914-04" />
-			</group>
 			<group targetFramework="netstandard2.0">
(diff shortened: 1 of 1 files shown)


codebude:
As described here https://github.com/codebude/QRCoder/issues/190#issuecomment-604363035, this is fixed by changes of the csproj-file structure (and the drop of the nuspec file).

PR #194, opened 2019-07-15
Refactor GetEncodingFromPlaintext to be more efficient. Decrease size…

csturm83:
Refactor ```GetEncodingFromPlaintext``` to be more efficient. Decrease size and need of static lookups.

_**Note:**_ The ```IsInRange``` method is a subset of .NET Core's [char.IsDigit() implementation](https://github.com/dotnet/coreclr/blob/11e150aab5be697d1d677a3084e66dbdab344d2a/src/System.Private.CoreLib/shared/System/Char.cs#L209-L218).


codebude:
Note: The following commit correlates with this merge: https://github.com/codebude/QRCoder/commit/1c4e203e52fbdfc6f0f6757264fb0843dddb73cd

PR #214, opened 2019-12-05
Upgrade to support .NET Core 3.1

Anduin2017:

* Upgrade to support .NET Core 3.1

**Test plan**
Sorry. Not tested.

**Closing issues**

None.

### DIFF
--- a/QRCoder/QRCoder.NETCore20.csproj
+++ b/QRCoder/QRCoder.NETCore20.csproj
@@ -12,7 +12,7 @@
 
   <PropertyGroup>
 	<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-	<PackageId>Anduin.QRCoder.NET</PackageId>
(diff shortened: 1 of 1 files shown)


codebude:
Won't merge this request, because it would drop NETSTANDARD2.0 as target and replace it with NETSTANDARD2.1. Instead I added NETSTANDARD2.1 as additional target: https://github.com/codebude/QRCoder/commit/83d4263a36424af73f4e29502b64a797c909e866

codebude:
Decided to rollback and to not add NETSTANDARD2.1 as target since 2.1 based apps should be able to use the 2.0 version of the QRCoder library. So there's no advantage to define 2.1 as additional target framework.

jnyrup:
@Anduin2017 
.NET Core 3.1 will currently consume the netstandard2.0 target.
If you're experiencing issues with that combination, please describe what issues you're having and why adding the extra target would solve it.

PR #224, opened 2020-02-08
Update ShadowSocksConfig PayloadGenerator

studentmain:

* [x] Support plugin field in shadowsocks config
* [x] Update encryption method

Shadowsocks introduced [plugin support](https://shadowsocks.org/en/spec/Plugin.html) in their [URL](https://shadowsocks.org/en/spec/SIP002-URI-Scheme.html) and QRCode. Also added few [new cipher](https://shadowsocks.org/en/spec/AEAD-Ciphers.html)

**Test plan**

added a test case from shadowsocks-windows

**Closing issues**

N/A

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -1991,38 +1991,51 @@ namespace QRCoder
 
         public class ShadowSocksConfig : Payload
         {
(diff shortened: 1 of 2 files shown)


codebude:
Thanks for your time and effort. 👍

PR #244, opened 2020-05-29
Named constructors for QRCoder.PayloadGenerator.SwissQrCode.Contact

inabout:
**Summary**

We are using QRCoder from C++/CLI. With the conformance improvements in Visual Studio 2019 version 16.6 (see https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements?view=vs-2019#default-argument-import-in-ccli), there is now an ambiguous call error when using the QRCoder.PayloadGenerator.SwissQrCode.Contact constructor for a combined address (four strings). The reason for this is that the other public constructor (for a structured address) takes also four strings and two more strings that default to null.

I added "named constructors" (static methods returning a Contact). The existing public constructors are still there (to avoid breaking changes), but marked as obsolete/deprecated.

**Test plan**

In the second commit, I adjusted all the (now obsolete/deprecated) Contact constructor usages in the tests and they are still all passing.

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -903,6 +903,7 @@ namespace QRCoder
                 /// <param name="country">Two-letter country code as defined in ISO 3166-1</param>
                 /// <param name="street">Streetname without house number</param>
                 /// <param name="houseNumber">House number</param>
+                [Obsolete("This constructor is deprecated. Use WithStructuredAddress instead.")]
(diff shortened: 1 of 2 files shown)


inabout:
@codebude I don't want to bother too much. Just asking whether I can do anything to facilitate reviewing/merging?

codebude:
Hi @inabout ,

thanks for your input and sorry for the delay. Spare time currently is kind of getting rare. ;-) 
Since you kept an eye on not breaking the code and at the same time did some improvement, I don't see any reason not to merge it.

inabout:
Great! Thanks, @codebude. Do you plan to release a new version on NuGet anytime soon?

PR #265, opened 2020-12-05
Reduce the number of SVG rectangles

CTSR:
This PR implements the following feature:
* Reduce the number of SVG rectangles by about 50%~60%


codebude:
Hi @CTSR thanks for your valuable input! This is a really good improvement for the SVG renderer.

PR #273, opened 2021-02-22
Simple demo

hong4rc:
This change makes the demo code simple


codebude:
Thanks for your support @hong4rc !

PR #280, opened 2021-03-29
fix CreateRoundedRectanglePath

afilippov1985:
Fixed to make the line start at the same point where the arc ends.


afilippov1985:
in fact, the lines are not needed at all

PR #322, opened 2021-09-09
WhatsAppPayload updated with the last click to chat feature and link …

NeyerOlmos:
**Summary**

* [ ] Bug 1: The WhatsApp Payload generates QR but it do not redirect to the WhatsApp App

1.- The WhatsApp Payload generates QR with link that have the old link structure to send automatically messages, in this pull request i updated the link generation structure and now it generate a QR correctly and when click open WhatsApp to send the message automatically

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -331,7 +331,10 @@ namespace QRCoder
             /// <summary>
             /// Let's you compose a WhatApp message and send it the receiver number.
             /// </summary>
-            /// <param name="number">Receiver phone number</param>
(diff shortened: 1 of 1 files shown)


codebude (changes requested):
I like to make QRCoder as user-friendly/easy to use as possible. The parameter comment, pointing out the exact number format is a good thing. 
Nevertheless I would love to see some additional logic to remove illegal/wrong symbols in the number. 

Do you mind adding in addition a Regex string replace that replaces all invalid chars from the number string?  You could use the follow regex for example: https://regex101.com/r/wcshJu/1
(Replace everything the regex matches with an empty string.)

codebude:
I'm closing this, since I merge your changes via an additional branch. See #332

Issue #375, opened 2022-01-12
QR Code unable to be scanned properly on iOS


hmireault-devo:
## Type of issue
[X] Bug  

## Expected Behavior

QR Codes generated with the same payload on version 1.4.2 and version 1.4.3 should result in the same image, and should be properly readable no matter the platform.

## Current Behavior
On QRCoder 1.4.3, codes generated with a certain payload are not able to be read properly on Microsoft Authenticator on iOS. On Android, there are no problems reading the QR code. Codes generated on version 1.4.2 of the library work properly no matter the platform.

## Possible Solution

I suspect the space character in the issuer or label is causing the issue, and it may be handled differently in version 1.4.3 compared to the previous version. When trying to reproduce the issue with an issuer and label without spaces, the issue does not occur.

## Steps to Reproduce

Sample code to generate the faulty QR code image:

```
QRCodeGenerator qrcodeGenerator = new QRCodeGenerator();
PayloadGenerator.OneTimePassword generator = new PayloadGenerator.OneTimePassword()
{
    Secret = "M54WC6TFJB2TQSCFGA2WOWLFIJCWOWSV",
    Issuer = "My Test Issuer",
    Label = "My Test Label"
};

QRCodeData qrcodeData = qrcodeGenerator.CreateQrCode(generator);
QRCode qrcode = new QRCode(qrcodeData);
this.pbQR.Image = qrcode.GetGraphic(5);
```

Generating the QR code image with version 1.4.2 will give the following:
![QRCoder - Test 1 4 2](https://user-images.githubusercontent.com/70523677/149198726-45f97096-3a00-436a-a069-f8da30193712.png)

Generating the QR code image with version 1.4.3 will give the following:
![QRCoder - Test 1 4 3](https://user-images.githubusercontent.com/70523677/149198757-545532d4-4a26-47ff-abd2-d67bf06b586e.png)

When using the Microsoft Authenticator app on iOS and scanning the first code, everything works.
When using the Microsoft Authenticator app on iOS and scanning the second code, the error "Invalid URL or Barcode" is shown.

When using the Microsoft Authenticator app on Android, both codes work.

## Your Environment

* Version used:  1.4.3
* Compiled from source or NuGet package?:  NuGet package
* Payload/Encoded text:  Described above in the code sample
* Used payload generator:  OneTimePassword
* Environment (.NET 3.5, .NET 4.X, .NETCore, ...):  .NET 4.8

craigpeterson-GAS:
[...]

PR #391, opened 2022-03-08
Ensure that an OTP's issuer and label values are escaped correctly

josh-:
## Summary

* [x] Creating an OTP with an issuer/label that contains a space or other restricted characters does not properly escape the issuer/label

### What existing problem does the pull request solve?**

Currently, if you create an OTP with an issuer that contains a space, for example:

```cs
var oneTimePassword = new OneTimePassword {
	Issuer = "Google Google",
	Label = "test@google.com",
	Secret = "pwq6 5q55"
};
```

and then call `ToString` on that `oneTimePassword`, that will return:

`otpauth://totp/Google Google:test@google.com?secret=pwq65q55&issuer=Google%20Google`

(note the first "Google Google" is not escaped, like the second one in the `issuer` parameter)

which then looks like this in Google Authenticator after scanning the resulting QR code:

File

This PR resolves this issue by correctly escaping both instances of the issuer in the URL, which then appears correctly in Google Authenticator:

`otpauth://totp/Google%20Google:test@google.com?secret=pwq65q55&issuer=Google%20Google`

## Test plan

Test cases have been added.

## Closing issues

N/A

### DIFF
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -2028,11 +2028,11 @@ namespace QRCoder
                     throw new Exception("Label must not have a ':'");
                 }
 
-                if (Label != null && Issuer != null)
+                if (Label != null && escapedIssuer != null)
                 {
                     label = escapedIssuer + ":" + Label;
                 }
-                else if (Issuer != null)
+                else if (escapedIssuer != null)
                 {
                     label = escapedIssuer;
                 }
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -2030,11 +2030,11 @@ namespace QRCoder
 
                 if (Label != null && Issuer != null)
                 {
-                    label = Issuer + ":" + Label;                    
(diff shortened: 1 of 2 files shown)


prvacy:
I discovered the same problem yesterday, thank you for the fix!

tom-156842:
Thanks for this fix, it addresses issue #375.

I suggest also escaping the `Label` value, since an [email address](https://en.wikipedia.org/wiki/Email_address#Local-part) can contain `/` and `?` characters.

josh-:
> Thanks for this fix, it addresses issue #375.
> 
> I suggest also escaping the `Label` value, since an [email address](https://en.wikipedia.org/wiki/Email_address#Local-part) can contain `/` and `?` characters.

Thanks @tom-156842, done 👍

hdocsek:
@codebude Do you any update on when this fix will be released?

doggy8088 (approved):
LGTM

@codebude

Shane32:
All changes here look good.  Generated QR codes scan properly with Google Authenticator.  Tests have already been added :+1:

PR #408, opened 2022-06-29
Update ArtQRCode.cs

Tragen:
Fixed dot size calculation where the size is divided instead of multiplied.


Shane32:
Previously the pixel size factor does not appear to have any effect.

With pixel size factor of 0.7: (default is 0.8)

Before:

test

After:

test

Demo of default size factor of 0.8:

test

It would seem that this PR should be merged, assuming this is the intent of the pixel size factor.  However, any user code will now generate with smaller dots than previously, as previously it acted as though pixelSizeFactor was 1.  Test hashes will need to be updated.

Shane32:
Closes:
- #463

QRCoderTests/ArtQRCodeRendererTests.cs
```diff
@@ -96,7 +96,7 @@ public void can_render_artqrcode_from_helper()
             var bmp = ArtQRCodeHelper.GetQRCode("A", 10, Color.Black, Color.White, Color.Transparent, QRCodeGenerator.ECCLevel.L);
 
             var result = HelperFunctions.BitmapToHash(bmp);
-            result.ShouldBe("57ecaa9bdeadcdcbeac8a19d734907ff");
+            result.ShouldBe("6c1bb8257349b29f76e50c3fa8abd9c9");
```
Shane32, line 99: Do you want to change the default size from 0.8 to 1.0 on the helper as well?  Then this hash would have remained the same.

Shane32: See ArtQRCode.cs line 286

Shane32 (approved):
Looks good to me

QRCoderTests/ArtQRCodeRendererTests.cs
codebude: I've overseen it. Was just on my way fixing it, when I saw your comment.

PR #430, opened 2022-11-02
Added Unicode block element support

gliljas:
## Summary
This PR implements support for rendering QR codes using the [block elements](https://en.wikipedia.org/wiki/Block_Elements) available in Unicode. Two lines can be rendered as one, using FULL BLOCK, UPPER HALF BLOCK and LOWER HALF BLOCK. While it would also be possible to use the quadrant characters, reducing the string length even further, they are very rarely implemented in fonts. 

The name Utf8QRCode is definitely up for discussion :)

### DIFF
--- a/QRCoder/Utf8QRCode.cs
+++ b/QRCoder/Utf8QRCode.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using static QRCoder.QRCodeGenerator;
+
+namespace QRCoder
+{
+    public class Utf8QRCode : AbstractQRCode, IDisposable
+    {
+        /// <summary>
+        /// Constructor without params to be used in COM Objects connections
+        /// </summary>
+        public Utf8QRCode() { }
+
+        public Utf8QRCode(QRCodeData data) : base(data) { }
+
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


gliljas:
I see now that this was also done in #384 . Didn't see it, since it was called ASCII, which it isn't :)

codebude:
Hi @gliljas ,

thanks for your PR and the work you put into it. I reviewed your PR as also the aforementioned PR #384 and decided for the latter at the end. 

Reason: I like the integration into the existing ASCIIQRCode more that having a another separate renderer with nearly the same use-case/output. Nevertheless your code was very inspiring and a took parts of your ideas into the codebase from #384. (E.g. the invert parameter as also the fixed size/capacity of the string builder.) 
I mentioned in the release notes and hope this is fine for you: https://github.com/codebude/QRCoder/wiki/Release-Notes#150

Regarding the naming "ASCIIQRCode" and your note (*it was called ASCII, which it isn't :)*) on it, I added an explanation in the wiki which reads out as follows:

> Trivia: The renderer is called Ascii renderer, although it uses characters from the Utf-8 character set by default. The idea behind this is that the renderer uses a text-based graphical representation that is reminiscent of classic Ascii art. Naming it Utf8QrCode was rejected in the design phase, as it would be difficult to deduce from this name what type of return value the renderer has.

I hope this helps to avoid confusion. :-)
Thanks again!

PR #438, opened 2022-12-24
Performance update on GenerateQrCode function

HolyOne:
Faster generation of QRCode using StringBuilder


gfoidl (commented):
The `StringBuilder` could be used in `GenerateQrCode` as well, to avoid the string-manipulations (and these allocations).

codebude:
Hi @HolyOne , hi @gfoidl Thanks for your support and pointing me onto this topic. I just re-fined your PR a little bit. I think by setting the capacity we can squeeze out even a little bit more performance. https://github.com/codebude/QRCoder/commit/0c54783b18f33637e5c4bf42d07edd9fa30da0f9

PR #447, opened 2023-02-27
url payload generator should detect all-caps protocol

mharen:
## Summary
Updates Url Payload Generator to better handle when a given url includes HTTP/HTTPS.

* [x] Bug: URL Payload Generator incorrectly prefixes `http://` to inputs that already include `HTTP://`

## Test plan
* unit test added to verify correct behavior
* make sure solution builds and tests run sucessfully


QRCoder/PayloadGenerator.cs
```diff
@@ -311,7 +311,7 @@ public class Url : Payload
             private readonly string url;
 
             /// <summary>
-            /// Generates a link. If not given, http/https protocol will be added.
+            /// Generates a link. If protocol not given, http protocol will be added.
```
Shane32, line 314:
```suggestion
            /// Generates a link. If the protocol is not specified, the http protocol will be added.
```

Shane32:
The changes here do not fail any tests.  I suggest merging the PR.

Shane32 (approved):

mharen:
Thanks!

codebude:
Hi @mharen , hi @Shane32 ,

thanks for your help. The change fixes the problem, thus I merged it. :-)
In future I propably will replace it by a Regex, because I think the startswith solution is somehow weak. Think about an input like "http-service.com". It would be matched by the startswith and the generator wouldn't prefix it, even if the protocol is missing.
With an Regex we may avoid it: https://regex101.com/r/PbTCnq/1

PR #449, opened 2023-04-02
Implementation of ByteMatrixQRCode

Benjlet:
## Summary
This PR implements 'ByteMatrixQRCode', which allows a byte matrix representation of the generated QR Code data.

I found this to be a very portable format for writing pixels blocks to other image libraries, where these simply translate to a grid of 1s/0s. Originally I was using the ZXing library to achieve this and had to develop that conversion as a wrapper, so believe this is adding something useful.

Unit tests included and all passing for me - I ran this on .NET 6 in VS2022.

I did not update the main README in this request as this would likely be done after/if this is merged, so it can link to the Wiki.

Cheers - love the library.


codebude:
Hi @Benjlet ,

late (probably too late) reply, but doesn't this somehow duplicate the "raw data" export of QRCoder? The raw data export aims to export the "black/white" information in the most space efficient way to be shared with other applications.

Benjlet:
Hey @codebude, I didn't see that was already an option - I personally found this implementation easier to port to other libraries and thought I'd share it out.  I will set this to closed but feel free to open if you think there's some value in it being shared in future - cheers.

PR #407, opened 2022-06-23
Adds Support for ImageSharp and SkiaSharp using two independent projects for maximum compatibility

emorell96:
## Summary
This PR brings support for ImageSharp and SkiaSharp  in two different packages: QrCoder.ImageSharp and QrCoder.SkiaSharp

It brings three new classes: `ImageSharpQrCode` `ArtQrCode` (but for ImageSharp), and `SkiaSharpQrCode`.

Each of them adds support for their respective rendering libraries, and you can then use said libraries to save the qr code as a png or further manipulations since both ImageSharpQrCode and SkiaSharpQrCode return a native Image object from either library.

I've created a few extension methods for SkiaSharpQrCode that allows easy saving into a stream or directly into a file.

I am open to discussion on how more features could be added or the packages be structured.

This PR fixes/implements the following **bugs/features**:

* Adds support for ImageSharp
* Adds support for SkiaSharp
* Adds support for logo images being drawn on top of the QR codes and custom colors for both renders.
* Ports ArtQrCode to ImageSharp.
* No Breaking changes since they  are an additional module not necessarily to be packed with the original. 

### What existing problem does the pull request solve?**

The lack of customization of the QR codes in other platforms other than Windows. Indeed modern .NET6 depends a lot of other imaging libraries like SkiaSharp and ImageSharp to do any complex drawings since System.Drawing is lacking in a lot of other platforms other than Windows.

To solve this I have coded expansions of QRCoder that allow drawing of QR codes using said libraries. Thanks to these libraries image manipulation is abstracted and I can easily add things like custom colors for the QR code, logo images in the center of the QR code or on the bottom right, or I can port the ArtQrCode which used to depend on System.Drawing and made it unavailable in many platforms.

## Test plan

I added to new tests projects and placed them under a Solution folder called Tests. Each for each package. They allow generation of the QR codes and you can scan them with your phone to check that they scan properly.

## Closing issues

Fixes #397
Contributes towards [#242](https://github.com/codebude/QRCoder/issues/242)

Feel free to add ideas/proposals to further improve each package to get closer to a leaner QrCoder with more modern renderers. 

Finally, I haven't fully tested below NET5.0 since I do not have older frameworks on my dev machine. I've tested the new modules on NET6 and NET5.

PR #460, opened 2023-07-01
Add ImageSharp renderers for .net 6.0, 5.0 and standart 2.0

AntonKorn:
## Summary

This PR adds the ImageSharp cross platform library to the QRCoder. It helps to still use Base64QRCode and QRCode in linux or any other environment. New rederers are only included in .net 6.0, 5.0 and standart 2.0

In the project I am currently employed to we need the feature of generating Base64QRCode for different image types. And recently we added linux support. To still use this library we had to re-implement these classes by ourselves and with help of https://github.com/JPlenert/QRCoder-ImageSharp. Unfortunately we couldn't use the mentioned fork directly because of some internal concerns.

## Breaking changes
ImageType enum was moved into the separate file from below the Base64QRCode class. Otherwise, changes were made backwar-compatible. Only new classes were added, old ones were left as is.

## Test plan
Build QRCoderConsole project and go to the bin folder. Execute following command  in net45, net5.0, net5.0-windows, net6.0-windows and compare the results.
`.\QRCoderConsole.exe -p "Input text" -e Q -s 20 -o file.png`

Following tool can be used: https://www.diffchecker.com/image-compare/. net45 needs to be compared with all the others.

### DIFF
--- a/QRCoder/ImageSharp/Base64QRCode.cs
+++ b/QRCoder/ImageSharp/Base64QRCode.cs
@@ -1,15 +1,11 @@
-﻿#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0 || NET6_0_WINDOWS
-using System;
-using System.Drawing;
-using System.Drawing.Imaging;
+﻿using System;
 using System.IO;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.Formats;
 using static QRCoder.QRCodeGenerator;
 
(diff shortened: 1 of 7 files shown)


Shane32:
I wouldn't endorse adding a third party (non-MS) dependency to the base QRCoder NuGet package. I would rather see something like PR https://github.com/codebude/QRCoder/pull/407 where there are separate packages depending on the needs of the user.

PR #474, opened 2023-10-08
fix:修正在绘制自定义finderPatternImage时，同时绘制QuietZones会出现位置异常的BUG

koaqiu:
修正在绘制自定义finderPatternImage时，同时绘制QuietZones会出现位置异常的BUG

### DIFF
--- a/QRCoder/ArtQRCode.cs
+++ b/QRCoder/ArtQRCode.cs
@@ -120,9 +120,10 @@ namespace QRCoder
                         if (finderPatternImage != null)
                         {
                             var finderPatternSize = 7 * pixelsPerModule;
-                            graphics.DrawImage(finderPatternImage, new Rectangle(0, 0, finderPatternSize, finderPatternSize));
-                            graphics.DrawImage(finderPatternImage, new Rectangle(size - finderPatternSize, 0, finderPatternSize, finderPatternSize));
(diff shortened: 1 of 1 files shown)


Shane32:
English translation (via ChatGPT): When customizing the finderPatternImage, a bug occurs where drawing the QuietZones at the same time results in abnormal positioning.

Shane32:
`can_create_standard_qrcode_graphic_with_custom_finder` tests this scenario.  However, the finder image used is completely transparent.

Assuming we alter the test with this code:

```cs
            var finder = new Bitmap(70, 70);
            using (var g = Graphics.FromImage(finder))
            {
                g.FillRectangle(Brushes.Red, 0, 0, 70, 70);
            }
```

It would currently produce this image:

finder

With this PR (and the above alteration to the test) it will produce this image:

finder

I suggest the test be altered as shown above, the hash updated, and the PR merged.

codebude:
Can't checkout the source branch to edit/add test cases. I'll merge first and then sent in further commits to fix the test case.

codebude:
Updated the test case in https://github.com/codebude/QRCoder/commit/e094cd3bf59e345f4202ca6433452e9cf7b445df
Testrunner finished successfully: https://github.com/codebude/QRCoder/actions/runs/8790250819

PR #476, opened 2023-11-04
Fix ASCII QR codes being diagonally mirrored

Timwi:
The current implementation of ASCIIQRCode has the X and Y coordinates swapped, resulting in a QR code that is mirrored about the main diagonal. I am not sure why the resulting QR code still scans successfully, but for feature parity with the other formats I propose to correct this.


QRCoder/ASCIIQRCode.cs
```diff
@@ -52,7 +52,7 @@ public string[] GetLineByLineGraphic(int repeatPerModule, string darkColorString
                 var lineBuilder = new StringBuilder();
                 for (var x = 0; x < QrCodeData.ModuleMatrix.Count - quietZonesModifier; x++)
                 {
-                    var module = QrCodeData.ModuleMatrix[x + quietZonesOffset][((y + verticalNumberOfRepeats) / verticalNumberOfRepeats - 1)+quietZonesOffset];
+                    var module = QrCodeData.ModuleMatrix[((y + verticalNumberOfRepeats) / verticalNumberOfRepeats - 1)+quietZonesOffset][x + quietZonesOffset];
```
Shane32, line 55: This change looks correct to me.  The for loops are correct, with the `x` being the inner loop, but when pulling the data from the matrix, the x/y were swapped.  Compare to ArtQRCode which correctly has `var pixelIsDark = this.QrCodeData.ModuleMatrix[offset + y][offset + x];`

Test hashes need to be updated.

codebude: Thanks for the review. I fixed the test cases.

codebude:
@Timwi  thanks for your support!

PR #490, opened 2024-04-02
Allow SVG exporting under .NET 6 on Linux

Shane32:
## Summary

This PR allows SVGs to be created under .NET 6 on Linux.  The majority of the code is already compatible and does not use System.Drawing.  The only code that does is the constructor for SvgLogo.  This implementation simply removes only that constructor on .NET 6 on Linux, without removing any other code, allowing:
1. SVGs without an embedded logo may be created
2. The alternate SvgLogo constructor may be used, for SVG-based logos

This does not allow users to use PNG-based logos on .NET 6+ on Linux -- a 3rd constructor could be added for that scenario, either public or protected, where the user would supply the PNG data (generated from a third-party image library, for example).  This is solved by #491 which can be merged independently.

### What existing problem does the pull request solve?

This allows users of QRCoder on Linux to export SVG files

### What is unique about this proposal?

Does not introduce any new dependencies

## Test plan

The currently existing tests within the codebase have been updated to run on .NET 6 on Linux, excluding the one for the constructor that references `System.Drawing.Bitmap`.

### DIFF
--- a/QRCoderTests/SvgQRCodeRendererTests.cs
+++ b/QRCoderTests/SvgQRCodeRendererTests.cs
@@ -103,7 +103,7 @@ namespace QRCoderTests
             result.ShouldBe("4ab0417cc6127e347ca1b2322c49ed7d");
         }
 
-#if !NET6_0
+#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0 || NET6_0_WINDOWS
         [Fact]
         [Category("QRRenderer/SvgQRCode")]
         public void can_render_svg_qrcode_with_png_logo()
--- a/QRCoderTests/SvgQRCodeRendererTests.cs
+++ b/QRCoderTests/SvgQRCodeRendererTests.cs
@@ -1,4 +1,4 @@
-﻿#if !NETCOREAPP1_1 && !NET6_0
+﻿#if !NETCOREAPP1_1
 using System;
 using Xunit;
 using QRCoder;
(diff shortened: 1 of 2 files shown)


QRCoderTests/SvgQRCodeRendererTests.cs
```diff
@@ -103,6 +103,7 @@ public void can_render_svg_qrcode_without_quietzones_hex()
             result.ShouldBe("4ab0417cc6127e347ca1b2322c49ed7d");
         }
 
+#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0 || NET6_0_WINDOWS
```
Shane32, line 106: `#if !NET6_0` would exclude both .NET 6 on Linux and .NET 6 on Windows.  So the ifdef has been fixed to properly run on all applicable test targets

codebude:
Hi @Shane32 ,

thanks for your PR and your appreciation. I had a first look at your changes. It looks good to me so far. However, before I merge, I would like to run my test cases again. In addition, unfortunately the build and test pipeline (workflow.yml) is also broken. (I haven't made any progress on the project for far too long. :-( )

Long story short: I'm open to a merge, but I want to fix the workflow first.

Shane32:
@codebude Thanks for the update.  I've posted PR #492 to fix the workflows, which runs fine on my fork.  Of course I can't be sure that publishing will work properly, but it seemed to be a simple matter of bumping the workflow dependencies.

codebude:
Test cases ran without failures. Let's merge.

codebude:
Added .NET 6.0 support for SvgQRCode in the compatability matrix: https://github.com/codebude/QRCoder/wiki/Advanced-usage---QR-Code-renderers

Issue #455, opened 2023-06-02
Tests for old .NET Frameworks


MPrtenjak:
## Type of issue
[ ] Bug  
[X] Question (e.g. about handling/usage)  
[ ] Request for new feature/improvement  

## Expected Behavior
Why are tests for .NET 3.5 different than for the other platforms? I am asking that because tests are failing.

For instance, this test:
```
        public void can_create_standard_qrcode_graphic()
        {
            var gen = new QRCodeGenerator();
            var data = gen.CreateQrCode("This is a quick test! 123#?", QRCodeGenerator.ECCLevel.H);
            var bmp = new ArtQRCode(data).GetGraphic(10);

            var result = HelperFunctions.BitmapToHash(bmp);
#if NET35_OR_GREATER || NET40_OR_GREATER
            result.ShouldBe("11ebdda91b9632d016798cb6de2f5339");
#else
            result.ShouldBe("cb38c3156eaf13cdfba699bdafc3a84c");
#endif
        }
```

Even though I am running tests in .NET 3.5 the result is equal to "cb38c3156eaf13cdfba699bdafc3a84c", which sounds to me as correct. 
Why would a result of a function depend on the .NET version?

## Your Environment
I am using Visual Studio 2022. Just download from GitHub, open in VS 2022, and run tests.

Some test fails on .NET 3.5 and .NET 4.5

Thanks

codebude:
Then there's maybe a change in the .NET Framework implementation...

Background: 
The idea of the tests was to generate a PNG (because in opposite to JPEG it has lossless compression and the same input Image object should return the same PNG file/bytes every time) and build a hash of the resulting file. Thus it should be ensured, that the QRCode has to look exactly the same to make an positive assumption. 

When I wrote the tests the same Image object, saved to PNG file, created byte-wise different files, when running in .NET 3.5/.NET 4.X vs all other framework versions. Back then I checked the files in a hex editor and saw that the PNG headers, written bei .NET differed in dependency to the used framework version. 
Different PNG file headers resulted in different checksums. That's why I added the compiler flags/branches in the test case. 

If it returns the same checksum now, then there must have been a patch in the framework.

MPrtenjak:
Thanks, that is exactly what I thought.

Well, because I had to download .NET Framework 4.5 additionaly, maybe there is something wrong on my side.

Thanks for the answer and have a great day.

PR #492, opened 2024-04-02
Fix workflows

Shane32:
## Summary

* Fixes GitHub Action workflows.
* Removes ifdef for hashes on NETFX.  Not sure why they were there, but tests fail unless they are removed, both locally and within CI.

### DIFF
--- a/.github/workflows/wf-build-release-ci.yml
+++ b/.github/workflows/wf-build-release-ci.yml
@@ -10,11 +10,11 @@ jobs:
     env:
         ACTIONS_ALLOW_UNSECURE_COMMANDS: true
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
       name: Checkout Code
     
     - name: Install additional .NET SDKs
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v4
       with:
         dotnet-version: | 
           1.0.x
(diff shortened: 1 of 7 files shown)


.github/workflows/wf-build-release-ci.yml
```diff
@@ -10,11 +10,11 @@ jobs:
     env:
         ACTIONS_ALLOW_UNSECURE_COMMANDS: true
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
       name: Checkout Code
     
     - name: Install additional .NET SDKs
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v4
```
Shane32, line 17: Just FYI, I've had numerous problems with other GitHub workflows that are still using v1 of `actions/setup-dotnet`.

codebude: If it works, then let's use v4. I just ised v1 because it was the up-to-date version when setting up the workflow.

codebude:
> ## Summary
> * Removes ifdef for hashes on NETFX.  Not sure why they were there, but tests fail unless they are removed, both locally and within CI.

Full explanation why the ifdef was used: https://github.com/codebude/QRCoder/issues/455
But since I also can't replicate the behaviour (there might have been a change anywhere down the PNG encoder lane), I'm fine with removing them.

PR #493, opened 2024-04-02
Update PostscriptQRCode so it runs on Linux

Shane32:
## Summary

* Updates `PostscriptQRCode` so it runs on all platforms except .NET Standard 1.3
* Adds tests for `PostscriptQRCode`

### Notes

* Although I added tests to validate existing functionality, I did not attempt printing any of the postscript output to verify it myself.
* No functionality changes were made to `PostscriptQRCode`.

### DIFF
--- a/QRCoder/PostscriptQRCode.cs
+++ b/QRCoder/PostscriptQRCode.cs
@@ -1,4 +1,4 @@
-﻿#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0 || NET6_0_WINDOWS
+﻿#if !NETSTANDARD1_3
 using System;
 using System.Drawing;
 using static QRCoder.QRCodeGenerator;
@@ -6,9 +6,6 @@ using static QRCoder.QRCodeGenerator;
 namespace QRCoder
 {
 
-#if NET6_0_WINDOWS
-    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
-#endif
     public class PostscriptQRCode : AbstractQRCode, IDisposable
     {
         /// <summary>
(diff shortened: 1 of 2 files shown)


codebude:
I'm confused. How does this work? The PostscriptQRCode uses System.Drawing multiple times (Color, Size, ...), but per my knoweldge System.Drawing isn't available on non-Windows platforms any longer: https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only

Thus I thought that PostscriptQRCode can't work on non-Windows platforms.

Shane32:
Color and Size are located in the System.Drawing.Primitives assembly which is still supported. They are basically just structs and constants. 

See documentation here:

https://learn.microsoft.com/en-us/dotnet/api/system.drawing?view=net-8.0

> Some types in the System.Drawing namespace rely on [GDI+](https://learn.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-gdi-start), which is not supported in Windows services and ASP.NET Core and ASP.NET apps. These types are in the [System.Drawing.Common](https://www.nuget.org/packages/System.Drawing.Common/) NuGet package and include [System.Drawing.Bitmap](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.bitmap?view=net-8.0) and [System.Drawing.Font](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.font?view=net-8.0). However, primitive types in the namespace, such as [System.Drawing.Color](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.color?view=net-8.0), [System.Drawing.Size](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.size?view=net-8.0), [System.Drawing.Point](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.point?view=net-8.0), and [System.Drawing.Rectangle](https://learn.microsoft.com/en-us/dotnet/api/system.drawing.rectangle?view=net-8.0), can be used in any application.

PR #503, opened 2024-04-25
Cleanup ifdefs and revise hashing algorithm

Shane32:
Cleans up ifdefs

To the extent possible:
- `SYSTEM_DRAWING` is used for when System.Drawing.Common is available
- `NET6_0_OR_GREATER` is used to apply `SupportedOSPlatformAttribute` (maybe in future PR this is polyfilled)
- `!NETSTANDARD1_3` is used to exclude code that requires System.Drawing.Primitives

And for tests:
- `SYSTEM_DRAWING` is used for when System.Drawing.Common is available
- `TEST_XAML` is used for XAML tests
- `!NETCOREAPP1_1` is used to exclude tests for .NET Standard 1.1, and for hash checks for PNG tests

All other ifdefs are simplified and used as minimally as possible (e.g. `NETFRAMEWORK` vs `NET35 || NET40`)

Also, many tests were not performed on .NET 6, probably since the hashes don't match due to a new deflate algorithm.  Worse yet, they do not appear to be deterministic, so rerunning the test can produce a different hash.  I've changed the hashing algorithm to be deterministic based on the pixel values inside the `Bitmap`.  This only works on platforms with access to `Bitmap` so the PNG tests still have ifdefs when running on .NET Core 1.1.

### DIFF
--- a/QRCoder.Xaml/XamlQRCode.cs
+++ b/QRCoder.Xaml/XamlQRCode.cs
@@ -1,5 +1,4 @@
-﻿#if NETFRAMEWORK || NET5_0_WINDOWS || NET6_0_WINDOWS
-using System;
+﻿using System;
 using System.Windows;
 using System.Windows.Media;
 using static QRCoder.QRCodeGenerator;
@@ -93,5 +92,3 @@ namespace QRCoder.Xaml
         }
     }
 }
-
-#endif
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -2516,14 +2516,10 @@ namespace QRCoder
                 var cp = characterSet.ToString().Replace("_", "-");
                 var bytes = ToBytes();
 
-#if !NET35_OR_GREATER && !NETSTANDARD1_3_OR_GREATER
+#if NETCOREAPP
                 System.Text.Encoding.RegisterProvider(System.Text.CodePagesEncodingProvider.Instance);
 #endif
-#if NETSTANDARD1_3              
-                return Encoding.GetEncoding(cp).GetString(bytes,0,bytes.Length);
-#else
-                return Encoding.GetEncoding(cp).GetString(bytes);
-#endif
+                return Encoding.GetEncoding(cp).GetString(bytes, 0, bytes.Length);
             }
 
             /// <summary>
@@ -2536,7 +2532,7 @@ namespace QRCoder
             {
                 //Setup byte encoder
                 //Encode return string as byte[] with correct CharacterSet
-#if !NET35_OR_GREATER
+#if !NETFRAMEWORK   // -- why different ifdef than line 2519? --
                 Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
 #endif
                 var cp = this.characterSet.ToString().Replace("_", "-");
--- a/QRCoder/PayloadGenerator.cs
+++ b/QRCoder/PayloadGenerator.cs
@@ -2516,8 +2516,8 @@ namespace QRCoder
                 var cp = characterSet.ToString().Replace("_", "-");
                 var bytes = ToBytes();
 
-#if NETCOREAPP
-                System.Text.Encoding.RegisterProvider(System.Text.CodePagesEncodingProvider.Instance);
+#if !NETFRAMEWORK
+                Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
 #endif
                 return Encoding.GetEncoding(cp).GetString(bytes, 0, bytes.Length);
             }
@@ -2532,7 +2532,7 @@ namespace QRCoder
             {
                 //Setup byte encoder
                 //Encode return string as byte[] with correct CharacterSet
-#if !NETFRAMEWORK   // -- why different ifdef than line 2519? --
+#if !NETFRAMEWORK
                 Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
 #endif
                 var cp = this.characterSet.ToString().Replace("_", "-");
@@ -3111,7 +3111,7 @@ namespace QRCoder
             Encoding utf8 = Encoding.UTF8;
             byte[] utfBytes = utf8.GetBytes(message);
             byte[] isoBytes = Encoding.Convert(utf8, iso, utfBytes);
-            return iso.GetString(isoBytes, 0, isoBytes.Length);
+            return iso.GetString(isoBytes);
         }
 
         private static string EscapeInput(string inp, bool simple = false)
--- a/QRCoder/QRCodeData.cs
+++ b/QRCoder/QRCodeData.cs
@@ -22,7 +22,6 @@ namespace QRCoder
                 this.ModuleMatrix.Add(new BitArray(size));
         }
 
-        [Obsolete("This constructor will be removed in version 2.0.0.")]
         public QRCodeData(string pathToRawData, Compression compressMode) : this(File.ReadAllBytes(pathToRawData), compressMode)
         {
         }
--- a/QRCoder/QRCodeData.cs
+++ b/QRCoder/QRCodeData.cs
@@ -154,7 +154,6 @@ namespace QRCoder
             return rawData;
         }
 
-        [Obsolete("This constructor will be removed in version 2.0.0.")]
         public void SaveRawData(string filePath, Compression compressMode)
         {
             File.WriteAllBytes(filePath, GetRawData(compressMode));
--- a/QRCoder/QRCodeData.cs
+++ b/QRCoder/QRCodeData.cs
@@ -21,11 +21,12 @@ namespace QRCoder
             for (var i = 0; i < size; i++)
                 this.ModuleMatrix.Add(new BitArray(size));
         }
-#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0
+
+        [Obsolete("This constructor will be removed in version 2.0.0.")]
         public QRCodeData(string pathToRawData, Compression compressMode) : this(File.ReadAllBytes(pathToRawData), compressMode)
         {
         }
-#endif
+
         public QRCodeData(byte[] rawData, Compression compressMode)
         {
             var bytes = new List<byte>(rawData);
@@ -154,12 +155,11 @@ namespace QRCoder
             return rawData;
         }
 
-#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0
+        [Obsolete("This constructor will be removed in version 2.0.0.")]
         public void SaveRawData(string filePath, Compression compressMode)
         {
             File.WriteAllBytes(filePath, GetRawData(compressMode));
         }
-#endif
 
         public int Version { get; private set; }
 
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1107,11 +1107,7 @@ namespace QRCoder
             Encoding utf8 = Encoding.UTF8;
             byte[] utfBytes = utf8.GetBytes(value);
             byte[] isoBytes = Encoding.Convert(utf8, iso, utfBytes);
-#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0
             return iso.GetString(isoBytes);
-#else
-            return iso.GetString(isoBytes, 0, isoBytes.Length);
-#endif
         }
 
         private static string PlainTextToBinaryByte(string plainText, EciMode eciMode, bool utf8BOM, bool forceUtf8)
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1026,8 +1026,7 @@ namespace QRCoder
         private static bool IsValidISO(string input)
         {
             var bytes = Encoding.GetEncoding("ISO-8859-1").GetBytes(input);
-            //var result = Encoding.GetEncoding("ISO-8859-1").GetString(bytes);
-            var result = Encoding.GetEncoding("ISO-8859-1").GetString(bytes,0,bytes.Length);
+            var result = Encoding.GetEncoding("ISO-8859-1").GetString(bytes);
             return String.Equals(input, result);
         }
 
--- a/QRCoderTests/ArtQRCodeRendererTests.cs
+++ b/QRCoderTests/ArtQRCodeRendererTests.cs
@@ -1,4 +1,4 @@
-﻿#if !NETCOREAPP1_1 && !NET6_0
+﻿#if SYSTEM_DRAWING
 
 using Xunit;
 using QRCoder;
@@ -23,7 +23,7 @@ namespace QRCoderTests
             var bmp = new ArtQRCode(data).GetGraphic(10);
 
             var result = HelperFunctions.BitmapToHash(bmp);
-            result.ShouldBe("cb38c3156eaf13cdfba699bdafc3a84c");
+            result.ShouldBe("df510ce9feddc0dd8c23c54e700abbf0");
         }
 
         [Fact]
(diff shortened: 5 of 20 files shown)


Shane32:
lol - looks like I forgot to submit my self review last night!

QRCoder.Xaml/XamlQRCode.cs
```diff
@@ -1,5 +1,4 @@
-﻿#if NETFRAMEWORK || NET5_0_WINDOWS || NET6_0_WINDOWS
```
codebude, line 1: Can you explain, why you removed the constants without replacement? :-)

QRCoder/PayloadGenerator.cs
```diff
@@ -2536,7 +2532,7 @@ public byte[] ToBytes()
             {
                 //Setup byte encoder
                 //Encode return string as byte[] with correct CharacterSet
-#if !NET35_OR_GREATER
+#if !NETFRAMEWORK   // -- why different ifdef than line 2519? --
```
codebude, line 2535: > -- why different ifdef than line 2519? --

No clue, but I would say the latter is the correct one: https://nicolaiarocci.com/how-to-read-windows-1252-encoded-files-with-.netcore-and-.net5-/

QRCoder/QRCodeData.cs
```diff
@@ -21,11 +21,12 @@ public QRCodeData(int version)
             for (var i = 0; i < size; i++)
                 this.ModuleMatrix.Add(new BitArray(size));
         }
-#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0
+
+        [Obsolete("This constructor will be removed in version 2.0.0.")]
```
codebude, line 25: Marked as obsolete, because file handling should be done by the calling application? Or for what reason? (Don't get me wrong - no critique - I just want to understand the reason. :-) )

Shane32:
I'm heading out now but if you read my comments (which I forgot to post last night), I think it will answer some of your questions.  I should be able to answer any other questions by tomorrow.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1107,11 +1107,7 @@ private static string ConvertToIso8859(string value, string Iso = "ISO-8859-2")
             Encoding utf8 = Encoding.UTF8;
             byte[] utfBytes = utf8.GetBytes(value);
             byte[] isoBytes = Encoding.Convert(utf8, iso, utfBytes);
-#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0
             return iso.GetString(isoBytes);
```
codebude, line 1110: Before your simplified to `iso.GetString(isoBytes, 0, isoBytes.Length)` (the one with three overloads. See: https://github.com/codebude/QRCoder/pull/503/files#diff-6752f108aa953a2872b67db5e51503d099b3f0ddb0ae668433a2ba73e12eb13aL2522

I would say this should be done here, too, because when I implemented it, with .NET Core there was no GetString() overload with just one parameter. On the other hand, it seems like the test cases passed successfully - but how?

QRCoderTests/ArtQRCodeRendererTests.cs
```diff
@@ -23,7 +23,7 @@ public void can_create_standard_qrcode_graphic()
             var bmp = new ArtQRCode(data).GetGraphic(10);
 
             var result = HelperFunctions.BitmapToHash(bmp);
-            result.ShouldBe("cb38c3156eaf13cdfba699bdafc3a84c");
+            result.ShouldBe("df510ce9feddc0dd8c23c54e700abbf0");
```
codebude, line 26: Why did the hashes change? Can't see a change in the actual implementation of the renderer. Shouldn't they stay consistent?

QRCoderTests/Helpers/HelperFunctions.cs
```diff
@@ -2,10 +2,11 @@
 using System.Text;
 using System.IO;
 using System.Security.Cryptography;
+using System.Reflection;
 #if !NETCOREAPP1_1
```
codebude, line 6: `#if !NETCOREAPP1_1` is this a candidate for `#if SYSTEM_DRAWING`?

QRCoderTests/Helpers/HelperFunctions.cs
```diff
@@ -44,34 +45,53 @@ public static Bitmap BitmapSourceToBitmap(DrawingImage xamlImg)
...
         }
-#endif
 
 
 #if !NETCOREAPP1_1
-        public static string BitmapToHash(Bitmap bmp)
+        /// <summary>
+        /// Converts a bitmap to a hash string based on the pixel data
+        /// using a deterministic algorithm that ignores compression algorithm
+        /// differences across platforms.
```
codebude, line 69: ❤️❤️❤️ Wow! Thanks - that problem was on my bucket list for years. (My development machine gave different hashes as the Github Action runner.) Glad to see this hopefully solved.

QRCoderTests/Helpers/HelperFunctions.cs
```diff
@@ -44,34 +45,53 @@ public static Bitmap BitmapSourceToBitmap(DrawingImage xamlImg)
...
+            try
             {
-                bmp.Save(ms, System.Drawing.Imaging.ImageFormat.Png);
-                imgBytes = ms.ToArray();
-                ms.Dispose();
+                // Create an array to hold the bytes of the bitmap.
+                int bytes = Math.Abs(bitmapData.Stride) * bitmap.Height;
+                rgbValues = new byte[bytes];
+
+                // Copy the RGB values into the array.
```
codebude, line 84: RGB or RGBA values? Because some test cases rely on transparency.

QRCoderTests/Helpers/CategoryDiscoverer.cs
```diff
@@ -1,14 +1,14 @@
...
 using Xunit.Abstractions;
 #endif
 using Xunit.Sdk;
 
 namespace QRCoderTests.Helpers.XUnitExtenstions
 {
-#if NET35 || NET452
+#if NETFRAMEWORK
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
     public class CategoryAttribute : Attribute
```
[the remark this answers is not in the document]
codebude: This is legacy stuff and can be removed. In the past it helped Filtering the test cases in test explorer, but this functionalty doesn't work anymore. 

> While the C# syntax allows you to place a Category attribute on a SetUpFixture class, the attribute is ignored by NUnit and has no effect in current releases.
https://docs.nunit.org/articles/nunit/writing-tests/attributes/category.html

QRCoderTests/Helpers/HelperFunctions.cs
```diff
@@ -44,34 +45,53 @@ public static Bitmap BitmapSourceToBitmap(DrawingImage xamlImg)
...
 
 
 #if !NETCOREAPP1_1
-        public static string BitmapToHash(Bitmap bmp)
+        /// <summary>
+        /// Converts a bitmap to a hash string based on the pixel data
+        /// using a deterministic algorithm that ignores compression algorithm
+        /// differences across platforms.
+        /// </summary>
+        public static string BitmapToHash(Bitmap bitmap)
```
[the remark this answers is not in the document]
codebude: Hi @Shane32 I had a look and left a comment. All in all it looks great!

codebude:
> I'm heading out now but if you read my comments (which I forgot to post last night), I think it will answer some of your questions. I should be able to answer any other questions by tomorrow.

Sure, will check your comments. Thanks again! :-)

QRCoder.Xaml/XamlQRCode.cs
Shane32: This project only targets these tfms and so this condition is always true.

I didn't really look much at the other projects, to be honest.  I just focused on the main project and the test project.

QRCoder/QRCodeData.cs
Shane32: My goal was to eliminate ifdefs.  Here the method did not exist on the most recent platform (.NET 6).  If this was accidental, then we can just remove the ifdef and leave it available for all platforms.  On the other hand, if the intent was to remove this method from future versions, I suggest we just mark it obsolete and remove in v2 (with no ifdefs).

Personally, I am not much in favor of methods to read/write directly from the file system, favoring instead for methods that read/write to `Stream` instances.  I'm probably biased, as I'm typically writing server applications that is going to read/write data in an Azure blob or send over HTTP.   But whatever you think is fine with me.

So, just lmk if I should remove the `[Obsolete]` attribute or not.

QRCoderTests/ArtQRCodeRendererTests.cs
Shane32: All calls to `BitmapToHash` use the new hashing algorithm, which uses LockBits to copy the actual pixel values, rather than relying on the PNG encoder, so that the hash is deterministic regardless of platform.

QRCoderTests/Helpers/HelperFunctions.cs
Shane32: The LockBits call on line 75 specifies ARGB format, so it will capture the transparency channel.

QRCoder/QRCodeGenerator.cs
Shane32: So, you are suggesting that I change it to the one with three parameters?  Do you think the one with three parameters is better?  Or just for consistency?  Maybe we should change the other one to have just a single parameter, for simplicity?

According to the .NET documentation, the single-parameter version is supported by .NET Framework 1.1+, .NET Standard 1.3+, and all versions of .NET Core.  Perhaps you had previously targeted an earlier version of .NET Standard.  But now QRCoder's base .NET Standard target is 1.3.

https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=net-8.0#system-text-encoding-getstring(system-byte())

QRCoderTests/Helpers/HelperFunctions.cs
Shane32: Curiously, no.  The PNG tests for the `net6.0` tfm utilize System.Drawing to perform hashing of the generated images.

QRCoder/QRCodeGenerator.cs
Shane32: Unless you specify otherwise, I changed the other reference to use the overload with a single parameter.

QRCoder/PayloadGenerator.cs
Shane32: Updated

QRCoder/QRCodeGenerator.cs
codebude: Ok, then maybe the signatures have changed over time. Or I initially targeted another .net core version. Current solution is fine for me.

QRCoder/QRCodeData.cs
codebude: >  If this was accidental, then we can just remove the ifdef and leave it available for all platforms.

Yes, I would say that was accidental. 7 years ago, when I was implementing this, QRCoder targeted (besides others) also PCL (is this still a thing?). The PCL library back then had no File support if I remember correctly, thus I added `#if !PCL` as flag. See here: https://github.com/codebude/QRCoder/blame/a45f953488aa7bc9b95308ffeafa0cc4abefcc82/QRCoder/QRCodeData.cs#L24

5 years ago switching from PCL to .NET Standard the term was changed to `#if !NETSTANDARD1_1` (because File.WriteAllBytes is only available from .NET Standard 1.3 on):
https://github.com/codebude/QRCoder/blame/3a528d8234ed50e490d8f2176d309607679c88f7/QRCoder/QRCodeData.cs#L24

Later the condition was inverted from exclusion to inclusion `#if NETFRAMEWORK || NETSTANDARD2_0` and then changed multiple times `#if NETFRAMEWORK || NETSTANDARD2_0 || NETSTANDARD2_1` --> `#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0`.

So the inital intention was to suppress the constructor for environments where File.WriteAllBytes wasn't available. Since we do not target .NET Standard 1.1 any longer (https://github.com/codebude/QRCoder/blob/069e4a14e48a94f0d46a8449e31dcac5205665dc/QRCoder/QRCoder.csproj#L4) I would say just remove the ifdefs *and* the obsolete attribute.

> Personally, I am not much in favor of methods to read/write directly from the file system, favoring instead for methods that read/write to Stream instances.

Back then I just thought if someone doesn't want to deal with files but read from/to memory, the constructor taking byte[] and the method returning byte[] would be convenient enough.

codebude:
One more thing (besides https://github.com/codebude/QRCoder/pull/503#pullrequestreview-2023654646) before merging this, is the compatability matrix: https://github.com/codebude/QRCoder/wiki/Advanced-usage---QR-Code-renderers#2-overview-of-the-different-renderers

Has anything changed due to the updated ifdefs? (I would say no after reviewing the PR, but maybe I've overseen something.)

codebude:
Having a look onto the compatability matrix, I asked myself why e.g. PostscriptQRCode isn't compatible with .NET Standard 1.3. I checked the history/blames and saw that the exclusion ifdef intially was defined as `!PCL" and then `!NETSTANDARD1_1`. Since we're targeting 1.3 now, the exclusion is proably not necessary any longer. Same story for SvgQRCode.

I wish I had commented back then which feature/function was missing in PCL when I excluded it, so that I could check now if its available in .NET Standard 1.3. I guess I have to simply try out...

codebude:
> Having a look onto the compatability matrix, I asked myself why e.g. PostscriptQRCode isn't compatible with .NET Standard 1.3. I checked the history/blames and saw that the exclusion ifdef intially was defined as `!PCL" and then `!NETSTANDARD1_1`. Since we're targeting 1.3 now, the exclusion is proably not necessary any longer. Same story for SvgQRCode.
> 
> I wish I had commented back then which feature/function was missing in PCL when I excluded it, so that I could check now if its available in .NET Standard 1.3. I guess I have to simply try out...

Got it. It's System.Drawing not beeing availble for NET_STANDARD1_3. So ignore my last comment.

Shane32:
> Has anything changed due to the updated ifdefs?

No; this doesn't change the exposed API.

If you like, I can help write a test that produces a copy of the API so it is easy to tell when it changes.  So it will auto generate a file like this:

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.ApiTests/net60/GraphQL.approved.txt

In our case at GraphQL.NET, we tailored the routine to be specific per-tfm, which complicates the routine considerably but would be particularly useful here where you have different APIs for various tfms.  We also extended the routine to auto update when it fails, so it is simpler for devs.

https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.ApiTests/ApiApprovalTests.cs

QRCoder/QRCodeData.cs
Shane32:
```suggestion
```

QRCoder/QRCodeData.cs
```diff
@@ -154,12 +154,11 @@ public byte[] GetRawData(Compression compressMode)
             return rawData;
         }
 
-#if NETFRAMEWORK || NETSTANDARD2_0 || NET5_0
+        [Obsolete("This constructor will be removed in version 2.0.0.")]
```
Shane32, line 157:
```suggestion
```

Shane32:
> No; this doesn't change the exposed API.

Correction: with the exception of adding the methods that were accidentally missing from .NET 6

codebude (approved):

Shane32:
> If you like, I can help write a test that produces a copy of the API so it is easy to tell when it changes

- #504

codebude:
> If you like, I can help write a test that produces a copy of the API so it is easy to tell when it change

That looks great! I've seen that you already started in #504 - thanks! :-) 
Do you mind if I start to pack QRCoder for the release without waiting for 504?

Shane32:
> Do you mind if I start to pack QRCoder for the release without waiting for 504?

Certainly!  #504 won't change the compiled code.

PR #509, opened 2024-04-30
Performance improvements

Shane32:
# System

```
BenchmarkDotNet v0.13.12, Windows 10 (10.0.19045.4291/22H2/2022Update)
13th Gen Intel Core i9-13900K, 1 CPU, 32 logical and 24 physical cores
.NET SDK 8.0.202
  [Host]     : .NET 8.0.3 (8.0.324.11423), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.3 (8.0.324.11423), X64 RyuJIT AVX2
```

# Before

| Method           | Mean       | Error    | StdDev   | Gen0     | Gen1    | Allocated  |
|----------------- |-----------:|---------:|---------:|---------:|--------:|-----------:|
| CreateQRCode     |   245.4 us |  4.88 us |  5.80 us |  17.0898 |  0.2441 |  317.99 KB |
| CreateQRCodeLong | 7,531.1 us | 34.13 us | 31.92 us | 265.6250 | 15.6250 | 4920.13 KB |

# After

| Method           | Mean       | Error   | StdDev  | Gen0    | Allocated |
|----------------- |-----------:|--------:|--------:|--------:|----------:|
| CreateQRCode     |   100.7 us | 0.26 us | 0.22 us |  1.2207 |  23.18 KB |
| CreateQRCodeLong | 1,842.1 us | 3.19 us | 2.83 us | 23.4375 | 433.53 KB |

# Notes

Most of the performance improvements in this PR fall into a few categories:
- Use `BitArray` instead of `string`
- Eliminate concatenation and instead write/copy into buffers using offsets and counts
- Preallocation of arrays/lists to proper length
- Eliminate array allocation when unnecessary (e.g. don't call `AddRange`)
- Eliminate LINQ use
- Improve lookup tables / dictionaries

# Progress

So far about half the code has been converted to use `BitArray`.  There is still a notable amount of code that still uses strings that remains to be optimized.  Also the `Polynom` class/constructor is heavily used and should be optimized.


QRCoder/QRCodeGenerator.cs
```diff
@@ -139,20 +153,32 @@ public static QRCodeData GenerateQrCode(string plainText, ECCLevel eccLevel, boo
...
             if (eciMode != EciMode.Default)
             {
-                modeIndicator = DecToBin((int)EncodingMode.ECI, 4);
-                modeIndicator += DecToBin((int)eciMode, 8);
+                DecToBin((int)EncodingMode.ECI, 4, completeBitArray, ref completeBitArrayIndex);
+                DecToBin((int)eciMode, 8, completeBitArray, ref completeBitArrayIndex);
+            }
+            DecToBin((int)encoding, 4, completeBitArray, ref completeBitArrayIndex);
+            // write count indicator
+            DecToBin(dataInputLength, GetCountIndicatorLength(version, encoding), completeBitArray, ref completeBitArrayIndex);
```
codebude, line 174: Second call to GetCountIndicatorLength. (First call in line 161.) Shall we store the result of the first call and re-use the variable with the result here?

QRCoder/QRCodeGenerator.cs
```diff
@@ -254,82 +275,181 @@ private static QRCodeData GenerateQrCode(string bitString, ECCLevel eccLevel, in
...
-            for (var i = 0; i < fStr.Length; i++)
-                sbMask.Append((Convert.ToInt32(fStr[i]) ^ Convert.ToInt32(fStrMask[i])).ToString());
-            return sbMask.ToString();
         }
 
-        private static string GetVersionString(int version)
+        private static void ShiftTowardsBit0(BitArray fStrEcc, int num)
         {
-            var generator = "1111100100101";
+#if NETCOREAPP
```
codebude, line 397: Why is a distinction made here between NETCOREAPP and non-NETCOREAPP? Surely the RightShift/LeftShift methods are available in all Framework versions? https://learn.microsoft.com/de-de/dotnet/api/system.linq.expressions.expression.rightshift?view=net-8.0

QRCoder/QRCodeGenerator.cs
```diff
@@ -254,82 +275,181 @@ private static QRCodeData GenerateQrCode(string bitString, ECCLevel eccLevel, in
...
+            for (var i = 0; i < fStrEcc.Length - num; i++)
+                fStrEcc[i] = fStrEcc[i + num];
+            for (var i = fStrEcc.Length - num; i < fStrEcc.Length; i++)
+                fStrEcc[i] = false;
+#endif
+        }
+
+        private static void ShiftAwayFromBit0(BitArray fStrEcc, int num)
+        {
+#if NETCOREAPP
```
codebude, line 409: ditto

QRCoder/QRCodeGenerator.cs
Shane32: No, just .NET Standard 2.1 / .NET Core 2.0 and newer -- see BitArray.LeftShift here:

https://learn.microsoft.com/en-us/dotnet/api/system.collections.bitarray.leftshift?view=net-8.0

QRCoder/QRCodeGenerator.cs
Shane32: Good idea.  Done.

Shane32:
FYI, I have not yet gone back through all my changes to add comments and perform a self-review, etc.

QRCoder/QRCodeGenerator.cs
Shane32: By the way, `BitArray` appears to be heavily optimized within the .NET runtime (albeit for longer bit runs):

https://github.com/dotnet/runtime/blob/2ad47cd52ebe57c1b4c2e28f95cc4eed6fcb354d/src/libraries/System.Collections/src/System/Collections/BitArray.cs#L664

One nice feature is that it is resizable like a `List<T>`.  However, `BitArray` is an allocation, and it's underlying `int[]` array is an allocation also.  So after this PR, I may review again and see if `BitArray` itself can be eliminated anywhere.  For example, `GetVersionString` and `GetFormatString` always manipulate and return bit arrays that are less than 32 bits.  The math could be done in an `int`, eliminating a couple allocations.  But I'm not planning to remove every allocation necessarily.  It's more important to get the recursive math optimized.

Shane32:
By the way, I plan to attempt using ArrayPool for further optimizations but not in this PR.

Shane32:
@codebude I think this PR has enough changes.  The GitHub comparison should be reasonable enough to follow.  I would like to follow up with some other PRs:
- Improve readability
- Fix bug (?) with string handling
- Utilize ArrayPool and other memory optimizations

Shane32:
The benchmarks are updated in the PR description, showing 60-75% memory reduction and 100-400% speed increase.

Most likely further PRs will further reduce memory requirements, but not increase speed.

Shane32:
I'm not planning to re-engineer the mathematics, like to use vector math or anything.  While I'm sure it's possible, that's not my forte.

QRCoder/QRCodeGenerator.cs
```diff
@@ -802,20 +880,20 @@ public static int Score(ref QRCodeData qrCode)
                                 blackModules++;
 
                     var percent = (blackModules / (qrCode.ModuleMatrix.Count * qrCode.ModuleMatrix.Count)) * 100;
-                    var prevMultipleOf5 = Math.Abs((int) Math.Floor(percent/5)*5 - 50)/5;
-                    var nextMultipleOf5 = Math.Abs((int)Math.Floor(percent / 5) * 5 -45)/5;
-                    score4 = Math.Min(prevMultipleOf5, nextMultipleOf5)*10;
+                    var prevMultipleOf5 = Math.Abs((int)Math.Floor(percent / 5) * 5 - 50) / 5;
+                    var nextMultipleOf5 = Math.Abs((int)Math.Floor(percent / 5) * 5 - 45) / 5;
```
gfoidl, line 884: Instead of `/ 5` that could be done once in `percent` by `* 100 / 5`.

QRCoder/QRCodeGenerator.cs
```diff
@@ -802,20 +880,20 @@ public static int Score(ref QRCodeData qrCode)
...
+                    score4 = Math.Min(prevMultipleOf5, nextMultipleOf5) * 10;
 
                     return score1 + score2 + score3 + score4;
```
gfoidl, line 887:
```suggestion
                    return (score1 + score2) + (score3 + score4);
```
See [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGATCAFdgAGxgMAsuQAUA4WIa5IscmgZzR4pdBik1Ghdth19QzYuUwUASi49O1Hk6YB2CzvIMA1O9ilvvjCMPkZWANx23AC+kQyx9OpmChKksklalqqJ8hk6etnmoSYFhpY2sQ7OPMRu0qGeIZak1gF1lsGBNhGOPDHUUUA===).

With the paranthesis it breaks the register dependency on `xmm0`, and the cpu can utilize more parallelism.

QRCoder/QRCodeGenerator.cs
```diff
@@ -910,72 +989,81 @@ private static bool IsInRange(char c, char min, char max)
             return (uint)(c - min) <= (uint)(max - min);
         }
 
-        private static Polynom CalculateMessagePolynom(string bitString)
+        private static Polynom CalculateMessagePolynom(BitArray bitArray, int offset, int count)
         {
-            var messagePol = new Polynom();
-            for (var i = bitString.Length / 8 - 1; i >= 0; i--)
+            var messagePol = new Polynom(count / 8);
+            for (var i = count / 8 - 1; i >= 0; i--)
```
gfoidl, line 995: I'm not sure* if the JIT keeps `count / 8` in a register here (CSE).
So maybe make it explicit and use a separate variable for this?

\* for a simple loop it will, but here it's a `new` and a method call

QRCoder/QRCodeGenerator.cs
```diff
@@ -910,72 +989,81 @@ private static bool IsInRange(char c, char min, char max)
...
+            generatorPolynom.PolyItems.Add(new PolynomItem(0, 0));
+            var multiplierPolynom = new Polynom(numEccWords * 2);
             for (var i = 1; i <= numEccWords - 1; i++)
             {
-                var multiplierPolynom = new Polynom();
-                multiplierPolynom.PolyItems.AddRange(new[]{
-                   new PolynomItem(0,1),
-                new PolynomItem(i,0)
-                });
+                multiplierPolynom.PolyItems.Clear();
```
gfoidl, line 1012: Can you please add a comment why `Clear` is needed here? At first glance it looked like a fault to me, but actually it's to re-use `multiplierPolynom`. So maybe something like
```suggestion
                // Re-use multiplierPolynom, so clear it's content.
                multiplierPolynom.PolyItems.Clear();
```

QRCoder/QRCodeGenerator.cs
```diff
@@ -910,72 +989,81 @@ private static bool IsInRange(char c, char min, char max)
...
-        private static List<string> BinaryStringToBitBlockList(string bitString)
+        private static byte[] BinaryStringToBitBlockByteList(BitArray bitString, int offset, int count)
         {
             const int blockSize = 8;
-            var numberOfBlocks = (int)Math.Ceiling(bitString.Length / (double)blockSize);
-            var blocklist = new List<string>(numberOfBlocks);
-
-            for (int i = 0; i < bitString.Length; i += blockSize)
+            if (count % blockSize != 0)
+                throw new ArgumentException("Count must be a multiple of 8.", nameof(count));
```
gfoidl, line 1026: Move to a [throw-helper](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunHjykTFAwBiECAAoAlgDsMDSAFc5ASi49O1HtoUQZuebPnAANhDABrAMpSAXjAYBeBgA4A3Op0MpAMwYSlOQYAUgZTc2s7BwBCZwAGNS0vHQwACygIAHcGGRhsgEEoAHNFfBg5AFEEMBgABwwpPQkAIhYIZXl8RQMwh2wGLpMG2pMHCD8XADpmtBzsMvGA9tUVDyTuAF9PBm36JgEhBgAhbChpIMCMRO1NZN19QyDwyxt7J1c1u99/S5CwsxeUQYsQYCW2dwAKukshJVuCePDuPxBMRhFCMplYU4AHwMNIYnJ5BiFEplSrVOoNJqtZadbrGPoDRRDKQjMYTaazGTzGCLS4qOHrBhbagbIA) to make the happy-path shorter.

QRCoder/QRCodeGenerator.cs
```diff
@@ -910,72 +989,81 @@ private static bool IsInRange(char c, char min, char max)
...
+            var numberOfBlocks = count / blockSize;
```
gfoidl, line 1027:
```suggestion
            var numberOfBlocks = (int)((uint)count / blockSize);
```
Then it's just a bit-operation instead of the integer-division.

QRCoder/QRCodeGenerator.cs
```diff
@@ -910,72 +989,81 @@ private static bool IsInRange(char c, char min, char max)
...
+            var ret = 0;
+            for (int i = 0; i < count; i++)
+            {
+                ret ^= bitArray[offset + i] ? 1 << (count - i - 1) : 0;
+            }
+            return ret;
         }
 
-        private static string DecToBin(int decNum, int padLeftUpTo)
+        private static void DecToBin(int decNum, int bits, BitArray bitList, ref int index)
```
gfoidl, line 1058: Can this become
```suggestion
        private static void DecToBin(int decNum, int bits, BitArray bitList, int index)
        {
            // loop the same
            return index;
        }
```
Because of the `ref int index` on each increment the memory (stack) must be updated, otherwise it can be in a register.
```diff
- inc dword ptr [r9]
+ inc r9d
```

QRCoder/QRCodeGenerator.cs
```diff
@@ -1013,7 +1101,7 @@ private static int GetCountIndicatorLength(int version, EncodingMode encMode)
             }
         }
 
-        private static int GetDataLength(EncodingMode encoding, string plainText, string codedText, bool forceUtf8)
+        private static int GetDataLength(EncodingMode encoding, string plainText, BitArray codedText, bool forceUtf8)
         {
             return forceUtf8 || IsUtf8(encoding, plainText, forceUtf8) ? (codedText.Length / 8) : plainText.Length;
```
gfoidl, line 1106:
```suggestion
            return forceUtf8 || IsUtf8(encoding, plainText, forceUtf8) ? (int)((uint)codedText.Length / 8) : plainText.Length;
```
Bit-hack instead of division.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1041,61 +1129,64 @@ private static string PlainTextToBinary(string plainText, EncodingMode encMode,
...
-        private static string PlainTextToBinaryNumeric(string plainText)
+        private static readonly BitArray _emptyBitArray = new BitArray(0);
+
+        private static BitArray PlainTextToBinaryNumeric(string plainText)
         {
-            var codeText = string.Empty;
-            while (plainText.Length >= 3)
+            var bitArray = new BitArray(plainText.Length / 3 * 10 + (plainText.Length % 3 == 1 ? 4 : plainText.Length % 3 == 2 ? 7 : 0));
+            var index = 0;
+            for (int i = 0; (i + 2) < plainText.Length; i += 3)
```
gfoidl, line 1144:
```suggestion
            for (int i = 0; i < plainText.Length - 2; i += 3)
```
`i + 2` needs to be evaluated at each iteration, so `plainText.Length - 2` can be kept in a register.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1134,16 +1224,27 @@ private static string PlainTextToBinaryByte(string plainText, EciMode eciMode, b
...
-                codeText += DecToBin(b, 8);
-
-            return codeText;
+            return ToBitArray(codeBytes);
         }
 
+        private static BitArray ToBitArray(byte[] byteArray, int prefixZeros = 0)
+        {
+            // new BitArray(byteArray) is not used because it reverses the bit order within each byte
+            var bitArray = new BitArray(byteArray.Length * 8 + prefixZeros);
```
gfoidl, line 1233:
```suggestion
            var bitArray = new BitArray((int)((uint)byteArray.Length * 8) + prefixZeros);
```

QRCoder/QRCodeGenerator.cs
```diff
@@ -1134,16 +1224,27 @@ private static string PlainTextToBinaryByte(string plainText, EciMode eciMode, b
...
+            for (var i = 0; i < byteArray.Length; i++)
+            {
+                var byteVal = byteArray[i];
+                for (var j = 0; j < 8; j++)
+                {
+                    bitArray[i * 8 + j + prefixZeros] = (byteVal & (1 << (7 - j))) != 0;
```
gfoidl, line 1239:
```suggestion
                    bitArray[(int)((uint)i * 8) + j + prefixZeros] = (byteVal & (1 << (7 - j))) != 0;
```
Or use the bit-hack directly, to avoid the casts which make the code more unreadable. But w/ the bit-hack maybe a comment is needed?!

QRCoder/QRCodeGenerator.cs
Shane32: Currently the functionality relies on the caller's `index` variable being added.  We could move it out; then the caller would have to do this:

```cs
DecToBin(value, 8, bitArray, index);
index += 8;
DecToBin(value2, 8, bitArray, index);
index += 8;
```

versus

```cs
DecToBin(value, 8, bitArray, index);
DecToBin(value2, 8, bitArray, index);
```

Shane32:
@gfoidl Thanks for the review!  I merged as many of them as I could via GitHub.com, and I'll need to spend a little more time integrating the rest of your comments.

codebude:
@gfoidl Thank you very much for your commitment. This is really a top-level review! Great!

QRCoder/QRCodeGenerator.cs
gfoidl:
```c#
index = DecToBin(value, 8, bitArray, index);
```
`index` is passed in, worked on it, then returned with the proper value. That's best for perf.

If you want to keep the `ref int index` then use such an approach:
```c#
void Foo(ref int index)
{
    int idx = index;
    ...
    // use idx here, not index
    ...
    index = idx;
}
```
That way `idx` can be a register, and at the end of the method the value of `idx` is written back to the memory `index`.
But perf isn't as good, as `index` must be memory accessible, due to the `ref`, so can't be in a cpu-register.

QRCoder/QRCodeGenerator.cs
Shane32: I've changed it to all integer arithmetic:

```cs
//=== OLD ===
double blackModules;
var percent = (blackModules / (qrCode.ModuleMatrix.Count * qrCode.ModuleMatrix.Count)) * 100;
var prevMultipleOf5 = Math.Abs((int)Math.Floor(percent / 5) * 5 - 50) / 5;
var nextMultipleOf5 = Math.Abs((int)Math.Floor(percent / 5) * 5 - 45) / 5;

//=== NEW ===
int blackModules;
var percentDiv5 = blackModules * 20 / (qrCode.ModuleMatrix.Count * qrCode.ModuleMatrix.Count);
var prevMultipleOf5 = Math.Abs(percentDiv5 - 10);
var nextMultipleOf5 = Math.Abs(percentDiv5 - 9);
```

@gfoidl  Please help me verify this does exactly what it did before.

QRCoder/QRCodeGenerator.cs
```diff
@@ -35,6 +37,19 @@ public QRCodeGenerator()
         {
         }
 
+        static QRCodeGenerator()
```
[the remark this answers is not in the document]
Shane32: Done.  There already exists tests for the galois fields, and they still pass now with the pre-generated table.

QRCoder/QRCodeGenerator.cs
```diff
@@ -167,169 +194,239 @@ public static QRCodeData GenerateQrCode(byte[] binaryData, ECCLevel eccLevel)
...
+                        fStrEcc[1] = true;
+                        break;
+                    default: // M: 00
+                        break;
+                }
+                int indexTemp = 2;
+                DecToBin(maskVersion, 3, fStrEcc, ref indexTemp);
+            }
+
+            void TrimLeadingZeros()
```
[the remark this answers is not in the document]
Shane32: Done

QRCoder/QRCodeGenerator.cs
```diff
@@ -167,169 +194,239 @@ public static QRCodeData GenerateQrCode(byte[] binaryData, ECCLevel eccLevel)
...
-            while (vStrEcc.Length > 12)
+        private static readonly BitArray _getVersionGenerator = new BitArray(new bool[] { true, true, true, true, true, false, false, true, false, false, true, false, true });
+        private static BitArray GetVersionString(int version)
+        {
+            var vStr = new BitArray(18);
+            var index = 0;
+            DecToBin(version, 6, vStr, ref index);
+            var count = vStr.Length;
+            index = 0;
+            while (!vStr[index])
```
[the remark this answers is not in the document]
Shane32: Done (latter)

QRCoder/QRCodeGenerator.cs
```diff
@@ -167,169 +194,239 @@ public static QRCodeData GenerateQrCode(byte[] binaryData, ECCLevel eccLevel)
...
+                count--;
             }
-            vStrEcc = vStrEcc.PadLeft(12, '0');
-            vStr += vStrEcc;
-
+            while (count > 12)
+            {
+                for (var i = 0; i < _getVersionGenerator.Length; i++)
+                    vStr[index + i] ^= _getVersionGenerator[i];
+                while (!vStr[index])
```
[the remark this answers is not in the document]
Shane32: Done

QRCoder/QRCodeGenerator.cs
Shane32: Done

QRCoder/QRCodeGenerator.cs
Shane32: Done

QRCoder/QRCodeGenerator.cs
Shane32: Done.  Nicer to use too.

Shane32:
Benchmarks updated.  New test results shaved 10us from short test and 20us from long test.

Shane32:
Eliminating `foreach` across a `BitArray` and implementing `IEquatable` for `Point` additionally cut down memory requirements for small QR codes by 90%, and above 60% for large QR codes.

codebude:
> Benchmarks updated. New test results shaved 10us (10%) from short test and 20us (1%) from long test.

But memory usage improved massively. :-)

Shane32:
While all beneficial changes, the memory usage didn't change due to the arithmetic changes.  The memory usage changes were due to `foreach` across a `BitArray` internally cast a `bool` to an `object` causing a heap allocation.  Ditto for `List<Point>.Contains` due to `Point` not implementing `IEquatable`.  These were accessed across some 'hot' code, and so many unnecessary heap allocations were made.

QRCoder/QRCodeGenerator.cs
gfoidl: From strict math I believe there may be some off by 1 errors with this conversion.
But for the use case (blackModules <= qrCode.ModuleMatrix.Count -- is that always true?) it holds, which I verified by [this sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmABgFgAoUgRmuoDNoACACgEs8MXJ8eAvC1oBuXhH4sAPENpkyYgNSK+3AJTUA3tRa6WzKOy49gAGwCGYANYBZCABMcpmAGcWQhSzOXbDp6+khVQwlRW9rO0dnFw0qPRZtOPi9ADdzQwhTewAlVydBFgB5LLZw3yjXNHF+NREdZN00wzwYAHdcl3z3FgA5NtKLCL9oquDa+oaJ5I5Gdhb2vNMeAEIhTJzFjFiG3USd5KJaAE42ABIAIgAxcw5THFgWcx4AIRsEssj/FwBfR7x7FgAYQSwW+53GSX2ukOJ3WHXyEKhehhbHm8KWiKRLBRmIa3ym+KohOoKPO9ggLXBdSoJNoSCM3CqxjURRK5JwZhgXkG5S+TO41XUWim7M5LAADjAoGAYAKhAMfJ9oiwAPTsYIsABUgq2LO1cgUU2MEtgKRs+Q44uchUYAFZujYngALAB0AEFgC42Jx1I6MK7LqYINA2JLpbKeGrbXqWPa4LGyCyo9T4saWggMOalpbrXaHc73Z7vcy/QGgyGwzKBVGY3GWCho6rYym9FMiAB2dji01ZjA5mA221VdOZi1Wgd2iHEmh0hkYflbXr9Y0fYaVFjGsbCyEbgWViMAEQ4KXtQlXFTc2vITbYGu1YxbumN3ZgZrHudPLFLha9++4R5PFh4zkXFdx4Ede37Qd839H9QylKsMAAusjlAttO1DHt3wnIcWAg7DBynIA==) (note: you can change the `100` to higher numbers, but it takes longer then, so for posting the link I kept the smaller number).

If 
> blackModules <= qrCode.ModuleMatrix.Count -- is that always true?

is always true, then you've done a really nice transform 👍🏻

codebude: This must be always true. blackModules are defined as blocks in ModuleMatrix with bit 1/true. Thus blackModules is a subset of ModuleMatrix and so can't be bigger than ModuleMatrix.Count.

QRCoder/QRCodeGenerator.cs
```diff
@@ -354,19 +342,20 @@ void WriteEccLevelAndVersion()
...
                         break;
                 }
-                int indexTemp = 2;
-                DecToBin(maskVersion, 3, fStrEcc, ref indexTemp);
+                DecToBin(maskVersion, 3, fStrEcc, 2);
             }
+        }
 
-            void TrimLeadingZeros()
+#if NETCOREAPP
```
gfoidl, line 349: Why is the #ifdef needed? The attribute should be available on all targets.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1043,6 +1023,8 @@ private static byte[] BinaryStringToBitBlockByteList(BitArray bitString, int off
             }
 
             return blocklist;
+
+            void ThrowArgumentException(string paramName) => throw new ArgumentOutOfRangeException(paramName, "Count must be a multiple of 8.");
```
gfoidl, line 1027: This helper is used only in once place, so I would remove the parameter and hard-code it here.
That avoid the load of the constant string at the call-site, thus makes the happy-path smaller.

The method could then be named `ThrowCountNotMultipleOf8` to reflect that change.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1369,8 +1352,10 @@ private static int GetIntValFromAlphaExp(int exp)
         private static int GetAlphaExpFromIntVal(int intVal)
         {
             if (intVal == 0)
-                throw new ArgumentOutOfRangeException(nameof(intVal));
+                ThrowArgumentOutOfRangeException(nameof(intVal));
             return galoisFieldByIntegerValue[intVal];
+
+            void ThrowArgumentOutOfRangeException(string paramName) => throw new ArgumentOutOfRangeException(paramName);
```
gfoidl, line 1358:
```suggestion
            void ThrowIntValIs0() => throw new ArgumentOutOfRangeException(nameof(intVal));
```
?
(similar to above)

gfoidl (commented):
A few nits, otherwise LGTM 👍🏻

QRCoder/QRCodeGenerator.cs
Shane32: Well, I should have enabled it on .NET Standard (fixed).

It is not available on .NET Framework prior to 4.5, and QRCoder targets 3.5 and 4.0, not 4.5.  See docs:

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimploptions?view=netframework-4.0

QRCoder/QRCodeGenerator.cs
Shane32: Done

QRCoder/QRCodeGenerator.cs
gfoidl: Ah, I just looked at the `MethodImpl` itself, not at `AggressiveInlining`. Thanks!

codebude (approved):

codebude (approved):

gfoidl (approved):

codebude:
Merged. Thanks for your great work @Shane32 & @gfoidl ! 👏

PR #514, opened 2024-05-02
[WIP] Fix character encoding when EciMode is specified

### LINKED ISSUE #513: What is the purpose of `forceUtf8`?
## Question

When generating a QR code with `QRCodeGenerator.CreateQrCode`, what is the purpose of `forceUtf8` versus `eciMode: EciMode.Utf8`?

I would expect that specifying `EciMode.Utf8` would force the generated code to use the UTF-8 character set when encoding the data.  But in fact, it does so only when either (a) there are non-ISO-8859-1 characters present, or (b) `forceUtf8` is `true`.

Consider the following:

```
// produces an unreadable code:
var qrData = new QRCodeGenerator().CreateQrCode("https://en.wikipedia.org/wiki/È", ECCLevel.L, eciMode: EciMode.Utf8);
// works fine:
var qrData = new QRCodeGenerator().CreateQrCode("https://en.wikipedia.org/wiki/È", ECCLevel.L, forceUtf8: true, eciMode: EciMode.Utf8);

// whereas this produces identical QR codes:
var qrData = new QRCodeGenerator().CreateQrCode("https://en.wikipedia.org/wiki/🍕", ECCLevel.L, eciMode: EciMode.Utf8);
var qrData = new QRCodeGenerator().CreateQrCode("https://en.wikipedia.org/wiki/🍕", ECCLevel.L, forceUtf8: true, eciMode: EciMode.Utf8);
```

A similar phenomenon occurs when specifying `EciMode.Iso8859_2` with specific strings -- interestingly, `forceUtf8` must be `true` for the code to function correctly.

If we assume that this is a bug, and `EciMode.Utf8` should always encode with UTF-8, then what is the purpose for the `forceUtf8` argument?  My best guess is that it is some compatibility mode to use UTF-8 encoding when encoded with `EciMode.Default` (which normally encodes as ISO-8859-1).

## Suggestion

I suggest that:

1. When specifying a `EciMode` besides `EciMode.Default`, the text is always encoded in the specified encoding.
2. For `EciMode.Default` with `forceUtf8 == false`, it uses ISO-8859-1 (per spec, `EciMode.Iso8859_1` is default)
3. For `EciMode.Default` with `forceUtf8 == true`, it uses UTF-8 (against spec)

## Spec

From ISO spec page 20:

> The default interpretation for QR Code is ECI 000003 representing the ISO/IEC 8859-1 character set.

Shane32:
Closes #513


QRCoder/QRCodeGenerator.cs
```diff
@@ -1015,19 +1015,7 @@ private static int GetCountIndicatorLength(int version, EncodingMode encMode)
...
-        {
-            return (encoding == EncodingMode.Byte && (!IsValidISO(plainText) || forceUtf8));
-        }
-
-        private static bool IsValidISO(string input)
-        {
-            var bytes = Encoding.GetEncoding("ISO-8859-1").GetBytes(input);
-            var result = Encoding.GetEncoding("ISO-8859-1").GetString(bytes);
-            return String.Equals(input, result);
+            return encoding == EncodingMode.Byte ? (codedText.Length / 8) : plainText.Length;
```
gfoidl, line 1018:
```suggestion
            return encoding == EncodingMode.Byte ? (int)((uint)codedText.Length / 8) : plainText.Length;
```
do we care about such things here too?

Shane32: Yes; we certainly can optimize this PR also.

I need to come back to this PR; but for now I was leaving it alone until I'm finished with performance PRs.

codebude:
Hi @Shane32 ,  

codebude:
Closed in consultation with @Shane32

PR #519, opened 2024-05-06
Utilize ArrayPool for ECC calculations

Shane32:
Improves performance by calculating `generatorPolynom` only a single time and renting arrays from `ArrayPool`.  Basically I rewrote `Polynom` to function similar to `List<PolynomItem>` with methods such as `Add`, `Clear` and `Sort`.  It's a struct so it does not require a heap allocation itself (no change), and the underlying buffer is rented from the array pool.  Calling `Dispose` on the `Polynom` will return the array back to the array pool.

## Before

| Method           | Mean       | Error   | StdDev  | Gen0    | Allocated |
|----------------- |-----------:|--------:|--------:|--------:|----------:|
| CreateQRCode     |   107.2 us | 0.66 us | 0.62 us |  1.2207 |  23.18 KB |
| CreateQRCodeLong | 1,878.4 us | 4.68 us | 4.15 us | 23.4375 | 433.53 KB |

## After - via ArrayPool<T>.Shared

| Method           | Mean        | Error    | StdDev   | Gen0   | Allocated |
|----------------- |------------:|---------:|---------:|-------:|----------:|
| CreateQRCode     |    99.44 us | 0.506 us | 0.448 us | 0.7324 |  14.68 KB |
| CreateQRCodeLong | 1,637.04 us | 2.811 us | 2.347 us | 1.9531 |  45.75 KB |

## After - via custom array pool implementation (for use with .NET Framework)

| Method           | Mean        | Error    | StdDev   | Gen0   | Allocated |
|----------------- |------------:|---------:|---------:|-------:|----------:|
| CreateQRCode     |    98.51 us | 0.188 us | 0.157 us | 0.7324 |  14.68 KB |
| CreateQRCodeLong | 1,706.09 us | 2.382 us | 2.112 us | 1.9531 |  45.75 KB |

### DIFF
--- a/QRCoder/QRCodeGenerator.Polynom.cs
+++ b/QRCoder/QRCodeGenerator.Polynom.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Text;
 
 namespace QRCoder
@@ -21,7 +22,7 @@ namespace QRCoder
             public Polynom(int count)
             {
                 _length = 0;
-                _polyItems = Allocate(count);
+                _polyItems = RentArray(count);
             }
 
             /// <summary>
@@ -29,7 +30,7 @@ namespace QRCoder
             /// </summary>
             public void Add(PolynomItem item)
             {
-                EnsureCapacity(_length + 1);
+                AssertCapacity(_length + 1);
                 _polyItems[_length++] = item;
             }
 
@@ -53,16 +54,25 @@ namespace QRCoder
             public PolynomItem this[int index]
             {
                 get {
-                    if (index < 0 || index >= _length)
-                        throw new IndexOutOfRangeException();
+                    if ((uint)index >= _length)
+                        ThrowIndexOutOfRangeException();
                     return _polyItems[index];
                 }
                 set {
-                    if (index < 0 || index >= _length)
-                        throw new IndexOutOfRangeException();
+                    if ((uint)index >= _length)
+                        ThrowIndexOutOfRangeException();
                     _polyItems[index] = value;
                 }
             }
+
+#if NET6_0_OR_GREATER
+            [StackTraceHidden]
+#endif
+            private static void ThrowIndexOutOfRangeException()
+            {
+                throw new IndexOutOfRangeException();
+            }
+
 
             /// <summary>
             /// Gets the number of polynomial terms in the polynomial.
@@ -100,7 +110,9 @@ namespace QRCoder
                 if (comparer == null) throw new ArgumentNullException(nameof(comparer));
 
                 var items = _polyItems;
-                if (items == null || _length <= 1)
+                if (items == null) throw new ObjectDisposedException(nameof(Polynom));
+
+                if (_length <= 1)
                 {
                     return; // Nothing to sort if the list is empty or contains only one element
                 }
@@ -158,14 +170,14 @@ namespace QRCoder
             /// <inheritdoc/>
             public void Dispose()
             {
-                Free(_polyItems);
+                ReturnArray(_polyItems);
                 _polyItems = null;
             }
 
             /// <summary>
             /// Ensures that the polynomial has enough capacity to store the specified number of polynomial terms.
             /// </summary>
-            private void EnsureCapacity(int min)
+            private void AssertCapacity(int min)
             {
                 if (_polyItems.Length < min)
                 {
@@ -173,12 +185,15 @@ namespace QRCoder
                     ThrowNotSupportedException();
 
                     // Sample code for growing the capacity:
-                    //var newArray = Allocate(Math.Max(min - 1, 8) * 2); // Grow by 2x, but at least by 8
+                    //var newArray = RentArray(Math.Max(min - 1, 8) * 2); // Grow by 2x, but at least by 8
                     //Array.Copy(_polyItems, newArray, _length);
-                    //Free(_polyItems);
+                    //ReturnArray(_polyItems);
                     //_polyItems = newArray;
                 }
 
+#if NET6_0_OR_GREATER
+                [StackTraceHidden]
+#endif
                 void ThrowNotSupportedException()
                 {
                     throw new NotSupportedException("The polynomial capacity is fixed and cannot be increased.");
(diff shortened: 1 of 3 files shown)


Shane32:
I don't know why these tests pass.  `CalculateECCWords` mutates `generatorPolynom`.  I have to rework this PR.

Shane32:
Just for fun, I've added another benchmark for a 2600-character QR code and compared this PR against v1.5.1.  For these unreasonably-long QR codes, memory requirements have been reduced from 110 MB to 114 KB, a 99.9% reduction.  And there's still room for improvement.

## Version 1.5.1

| Method              | Mean        | Error     | StdDev    | Gen0      | Gen1     | Allocated    |
|-------------------- |------------:|----------:|----------:|----------:|---------:|-------------:|
| CreateQRCode        |    245.4 us |   2.36 us |   1.97 us |   17.0898 |        - |    317.99 KB |
| CreateQRCodeLong    |  7,492.1 us |  30.09 us |  26.67 us |  265.6250 |  15.6250 |   4920.13 KB |
| CreateQRCodeLongest | 88,024.3 us | 209.67 us | 185.86 us | 5833.3333 | 833.3333 | 110090.17 KB |

## This PR

| Method              | Mean         | Error     | StdDev    | Gen0   | Allocated | % RAM reduction |
|-------------------- |-------------:|----------:|----------:|-------:|----------:|----:|
| CreateQRCode        |     98.12 us |  0.213 us |  0.189 us | 0.7324 |  14.68 KB | 95% |
| CreateQRCodeLong    |  1,675.42 us |  1.812 us |  1.415 us | 1.9531 |  45.75 KB | 99% |
| CreateQRCodeLongest | 14,824.82 us | 48.627 us | 45.486 us |      - | 113.96 KB | 99.9% |

QRCoder/QRCodeGenerator.Polynom.cs
```diff
@@ -8,21 +9,131 @@ public partial class QRCodeGenerator
...
+                EnsureCapacity(_length + 1);
+                _polyItems[_length++] = item;
+            }
+
+            /// <summary>
+            /// Removes the polynomial term at the specified index.
+            /// </summary>
+            public void RemoveAt(int index)
+            {
+                if (index < 0 || index >= _length)
```
gfoidl, line 41: You can save one comparison (branch) with
```suggestion
                if ((uint)index >= (uint)_length)
```
If `index` is negative, then the `uint` cast will result in values `> int.MaxValue`, so effectively doing the same, but w/ less work.

QRCoder/QRCodeGenerator.Polynom.cs
```diff
@@ -8,21 +9,131 @@ public partial class QRCodeGenerator
...
+                    throw new IndexOutOfRangeException();
```
gfoidl, line 42: Perf-wise it's better to use a throw-helper, like
```c#
if (...) ThrowIndexOutOfRange();

[StackTraceHidden]
static void ThrowIndexOutOfRange() => throw new IndexOutOfRangeException();
```
So the happy-path of the method remains short and doesn't contain exception logic (which is heavy). Instead it's separated out to a separate method.
Starting w/ .NET 3.1 (I think) the JIT can see that this method won't return (as it throws), so  the JIT places these method into a cold-region.

QRCoder/QRCodeGenerator.Polynom.cs
```diff
@@ -43,6 +154,134 @@ public override string ToString()
...
+                {
+                    throw new NotSupportedException("The polynomial capacity is fixed and cannot be increased.");
+                }
+            }
+
+#if NETCOREAPP
+            /// <summary>
+            /// Allocates memory for the polynomial terms.
+            /// </summary>
+            private static PolynomItem[] Allocate(int count)
```
gfoidl, line 192: Nitpicking: the name `Allocate` is a bit misleading, as amortized no allocation will happen. Maybe call it `GetPolynomItems`, `Rent`, ...

Hm, but as in general memory allocation can also be thought of as renting from the virtual memory, maybe the name is still correct. In .NET-terms for allocation I think about that the GC hands me out a new created object...

QRCoder/QRCodeGenerator.Polynom.cs
```diff
@@ -8,21 +9,131 @@ public partial class QRCodeGenerator
...
+            {
+                get {
+                    if (index < 0 || index >= _length)
+                        throw new IndexOutOfRangeException();
+                    return _polyItems[index];
+                }
+                set {
+                    if (index < 0 || index >= _length)
+                        throw new IndexOutOfRangeException();
+                    _polyItems[index] = value;
```
gfoidl, line 63: W/ the `uint`-trick from above the bound-checks can be elided by the JIT here too.

QRCoder/QRCodeGenerator.Polynom.cs
```diff
@@ -43,6 +154,134 @@ public override string ToString()
...
+            public void Dispose()
+            {
+                Free(_polyItems);
+                _polyItems = null;
+            }
+
+            /// <summary>
+            /// Ensures that the polynomial has enough capacity to store the specified number of polynomial terms.
+            /// </summary>
+            private void EnsureCapacity(int min)
```
gfoidl, line 168: This method (right now, w/ the commented out part) doesn't "ensure" the capacity, rather it
```suggestion
            private void AssertCapacity(int min)
```
😉.

Shane32:
Thanks @gfoidl !  I'll merge in your suggestions here later tonight.

QRCoder/QRCodeGenerator.Polynom.cs
Shane32: Done

QRCoder/QRCodeGenerator.Polynom.cs
Shane32: Done - RentArray & ReturnArray

QRCoder/QRCodeGenerator.Polynom.cs
Shane32: Done

QRCoder/QRCodeGenerator.Polynom.cs
Shane32: Done

PR #521, opened 2024-05-07
Add string optimizations

Shane32:
This PR prevents `PlainTextToBinaryByte` from encoding the provided string back and forth to the target encoding multiple times.  It also caches the encoder for future use, as it is designed to be a singleton.  On the .NET Core 5+ TFMs the encoding uses a stack-allocated byte array, reducing heap allocations and memory requirements.

Add tests for UTF8 with and without BOM.

No changes to logic; produced QR codes should be identical to previously for any input.


QRCoder/QRCodeGenerator.cs
```diff
@@ -745,14 +745,22 @@ bool IsUtf8()
...
 
+        private static readonly Encoding _iso88591ExceptionFallback = Encoding.GetEncoding(28591, new EncoderExceptionFallback(), new DecoderExceptionFallback()); // ISO-8859-1
         /// <summary>
         /// Checks if the given string can be accurately represented and retrieved in ISO-8859-1 encoding.
         /// </summary>
         private static bool IsValidISO(string input)
         {
-            var bytes = Encoding.GetEncoding("ISO-8859-1").GetBytes(input);
-            var result = Encoding.GetEncoding("ISO-8859-1").GetString(bytes);
-            return String.Equals(input, result);
```
Shane32, line 755: Here we eliminated conversion from UTF-16 to ISO-8859-1 and then back again (two conversions).

QRCoder/QRCodeGenerator.cs
```diff
@@ -866,18 +874,13 @@ private static BitArray PlainTextToBinaryAlphanumeric(string plainText)
...
-        /// specified encoding (default of ISO-8859-2) with a replacement character.
-        /// </summary>
-        private static string ConvertToIso8859(string value, string Iso = "ISO-8859-2")
-        {
-            Encoding iso = Encoding.GetEncoding(Iso);
-            Encoding utf8 = Encoding.UTF8;
-            byte[] utfBytes = utf8.GetBytes(value);
-            byte[] isoBytes = Encoding.Convert(utf8, iso, utfBytes);
-            return iso.GetString(isoBytes);
-        }
```
Shane32, line 880: Every time this method was called:
1. The string was converted from UTF-16 to UTF-8
2. The string was converted from UTF-8 to the target encoding
3. The string was converted from the target encoding back to UTF-16

Ultimately it did not do anything besides remove characters from the string that could not be represented in the target encoding

QRCoder/QRCodeGenerator.cs
```diff
@@ -894,35 +897,70 @@ private static string ConvertToIso8859(string value, string Iso = "ISO-8859-2")
...
                         break;
                     case EciMode.Iso8859_2:
+                        // Note: ISO-8859-2 is not natively supported on .NET Core
+                        //
+                        // Users must install the System.Text.Encoding.CodePages package and call Encoding.RegisterProvider(CodePagesEncodingProvider.Instance)
+                        // before using this encoding mode.
+                        if (_iso8859_2 == null)
+                            _iso8859_2 = Encoding.GetEncoding(28592); // ISO-8859-2
                         // Convert text to ISO-8859-2 and encode.
-                        codeBytes = Encoding.GetEncoding("ISO-8859-2").GetBytes(ConvertToIso8859(plainText, "ISO-8859-2"));
```
Shane32, line 913: The `Encoding.GetBytes` method would remove characters that do not fit in the target encoding anyway, so the call to `ConvertToIso8859` was redundant.

QRCoder/QRCodeGenerator.cs
```diff
@@ -866,18 +874,13 @@ private static BitArray PlainTextToBinaryAlphanumeric(string plainText)
...
-            byte[] isoBytes = Encoding.Convert(utf8, iso, utfBytes);
-            return iso.GetString(isoBytes);
-        }
+        private static readonly Encoding _iso8859_1 =
+#if NET5_0_OR_GREATER
+            Encoding.Latin1;
+#else
+            Encoding.GetEncoding(28591); // ISO-8859-1
+#endif
+        private static Encoding _iso8859_2;
```
Shane32, line 883: On .NET 5 and newer there is a static property available for ISO-8859-1.  Regardless, for all platforms, ISO-8859-1 is guaranteed to be available.  However, requesting ISO-8859-2 will throw an exception on Linux unless the code page is installed, so it is lazy loaded and cached.  This represents no change to behavior, as attempting to use ISO-8859-2 with v1.5.1 on Linux currently throws an exception (unless the code page provider is installed).

QRCoder/QRCodeGenerator.cs
```diff
@@ -745,14 +745,22 @@ bool IsUtf8()
...
-            var bytes = Encoding.GetEncoding("ISO-8859-1").GetBytes(input);
-            var result = Encoding.GetEncoding("ISO-8859-1").GetString(bytes);
-            return String.Equals(input, result);
+            // No heap allocations if the string is ISO-8859-1
+            try
+            {
+                _ = _iso88591ExceptionFallback.GetByteCount(input);
+                return true;
+            }
+            catch (EncoderFallbackException) // The exception is a heap allocation and not ideal
```
gfoidl, line 760: FYI: starting w/ .NET 8 one could use [`Encoding.TryGetBytes`](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.trygetbytes?view=net-8.0) for this to avoid the slow exception path.

(I know that ATM there's no target for .NET 8 or newer in this project)

QRCoder/QRCodeGenerator.cs
```diff
@@ -894,35 +897,69 @@ private static string ConvertToIso8859(string value, string Iso = "ISO-8859-2")
...
+                        targetEncoding = Encoding.UTF8;
                         break;
                 }
             }
 
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
+            // We can use stackalloc for small arrays to prevent heap allocations
+            // Note that all QR codes should fit within 3000 bytes, so this code should never trigger a heap allocation unless an exception will be thrown anyway.
+            int count = targetEncoding.GetByteCount(plainText);
+            Span<byte> codeBytes = count < 3000 ? stackalloc byte[count] : new byte[count];
```
gfoidl, line 942: The constant should be a multiple of 2, in order to create better machine code.

The stack-space must be zoroed*, and cpu registers, etc. are all multiple of 2.

\* except `[SkipLocalsInit]` is applied, then the zeroing can be omitted by the JIT
Note: if this attribute isn't available at the target framework, one can just provide a simple shim for it and the JIT will still recognize it.

Shane32: I'm not sure how this applies.  `3000` is used as a threshold to switch between `stackalloc` and `new byte[]`, and as described in the comments, it should never hit `new byte[]` under normal circumstances.

QRCoder/QRCodeGenerator.cs
Shane32: `Encoding.TryGetBytes`, where available, only is useful if it is determined that the buffer is too small.  So potentially it could eliminate `GetByteCount`, but it will not avoid the exception path.

This method should be eliminated in a separate PR anyway - see #514 .  In this PR the logic is unchanged so the method is necessary.

QRCoder/QRCodeGenerator.cs
gfoidl: The JIT isn't allowed to round up here, as it could change the behavior. Any optimizations must not change the visible behavior of the (sequential) program.

BTW: maybe 3000 bytes is a bit too big for safely stack allocating? Regular (thread) stack-sizes are big enough, but that method could be deeply nested where methods up the stack also could have larger amount of stack-space, so potentially throwing StackOverflowException? I'd cap it by some smaller values like 512, then rent from the array-pool if sizes > 512 are frequent.

For rounding up that would be 3000 -> 4096, which may be quite a difference.

But now I see that my comment isn't really complete ('cause it's too obvious for me), sorry for that.
The essence is that the value given to `stackalloc` is a constant, that is a multiple of two and not any variable length.
Perf-wise it's better to do `stackalloc byte[ConstantAsMultipleOfTwo].Slice(0, count)` than just `stackalloc byte[count]`.
That's what I tried to say with my previous comment.

The ideal pattern if count is most likely below the threshold is
```c#
Span<byte> buffer = count <= MaxStackSizeInBytes
    ? stackalloc byte[MaxStackSizeInBytes].Slice(0, count)   // Slice can be omitted when the buffer doesn't need to be exact size
    : new byte[count];
```
or w/ renting from the array-pool when larger sizes are equally likely
```c#
const int MaxStackSizeInBytes = 512;   // or whatever is a good choice here (128, 256), as long as it's a multiple of 2

byte[]? bufferFromPool = null;
Span<byte> buffer = (count <= MaxStackSizeInBytes
    ? stackalloc byte[MaxStackSizeInBytes]
    : bufferFromPool = ArrayPool<byte>.Shared.Rent(count))
    .Slice(0, count);

// Use buffer

if (bufferFromPool is not null)
{
    ArrayPool<byte>.Shared.Return(bufferFromPool);
}
```
Note: the array returned from the array-pool may also be longer than the length requested, thus it's sliced to the correct size -- except the buffer can be bigger than needed.

codebude:
Hello @gfoidl , hello @Shane32 - I would like to take this opportunity to say once again that I am very impressed by your suggestions and your discussion. To be honest, your optimizations reach a level that exceeds my knowledge of C#/.NET/JIT. Nevertheless, I think I vaguely understand what you are discussing and I think I have learned a thing or two. Thank you very much for your efforts. I bow my head!

QRCoder/QRCodeGenerator.cs
Shane32: Updated to suggestion

PR #524, opened 2024-05-14
Eliminate use of LINQ within GetVersion

Shane32:
Minor performance tweak to eliminate allocations within GetVersion (assuming no exception is thrown).

### DIFF
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -524,23 +524,21 @@ namespace QRCoder
         /// <returns>The minimum version of the QR code that can accommodate the given data and settings.</returns>
         private static int GetVersion(int length, EncodingMode encMode, ECCLevel eccLevel)
         {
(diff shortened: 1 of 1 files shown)


codebude (approved):

codebude:
Do you plan any further performance optimizations? Otherwise I would release & push the current state as 1.5.2. All those performance optimization already done are worth a release :-)

Shane32:
If I have time, I'd like to:
- use some kind of ArrayPool buffer(s) for the codeword ECC blocks
- make a custom BitArray that uses ArrayPool internally for two larger allocations which are internally freed
- decide if there should be any changes (or bug fixes) for strings - see #514
- review additionally targeting .NET Standard 2.1, as there would be native `Span<>` and ArrayPool support for older frameworks, I believe

I have nothing against releasing 1.5.2 at any time you wish -- the performance optimizations have no effect on the exposed API or its use, and any future performance enhancements will be minor in comparison.

Issue #512, opened 2024-05-01
🆕 Planning QRCoder 2.0 - let's do it together


codebude:
This issue is intended to initiate version 2.0 of the QRCoder. I would like to see an active discussion here so that we can shape the implementation of version 2.0 together. Feedback from the community is very important to me.

### Why QRCoder 2.0?

The current version of the QRCoder is essentially still based on the first developments from 2013. Although the QRCoder has been repeatedly adapted to the times, large parts are still based on System.Drawing.Common, which since .NET 6.0 has only been supported under Windows.
Since then, this has often led to confusion, as parts of the QRcoder have only been available under Windows since .NET 6.0, which despite I wrote an [Wiki entry](https://github.com/codebude/QRCoder/wiki/Advanced-usage---QR-Code-renderers#2-overview-of-the-different-renderers) is far from obvious to all users.

Although there have already been some PRs and suggestions for version 2.0 in the past, I have shied away from adopting the vision of individual contributors without consulting the community. I want to make up for this with this issue.

Version 2.0 should first and foremost provide clarity and create a consistent interface that is available on as many platforms as possible. Since this most likely means turning away from System.Drawing(.Common) and thus requires breaking changes, this release will be merged into a new major version (2.0).

### Requirements for QRCoder 2.0

- The QRCoder should work on as many platforms as possible (at least near full support under Windows and Linux)
- If a component is only available on one platform, this should be clearly recognizable and comprehensible for the user

### What needs to be clarified?

In order for the implementation of version 2.0 to be a success (both in terms of the result and the path to it), a few decisions need to be made in advance. **Here I am hoping for feedback from you, the community.**
*(If you give feedback on the points, please include the number of the point so that everyone knows what you are referring to)*.

1. **How do we split the QRCoder? (NuGet packages):** Do we leave only the QR code generator logic in the main library (*QRCoder*) and make a separate package for each renderer (e.g. *QRCoder.Svg*, *QRCoder.Png*, *QRCoder.Ascii*, etc.)? Or do we want the generator logic + all renderers that work on all platforms without dependencies in the main package (*QRcoder*) and create only NuGet package per renderer with ext. dependencies? (e.g. *QRCoder.SystemDrawing*, *QRCoder.ImageSharp*, ...) Should the payload generators also be in their own NuGet package?

[...]

PR #526, opened 2024-05-18
ECC code mismatch detection

Shane32:
This pull request enhances the QRCoder library by allowing the payload to specify an "unspecified" error correction level (ECC) rather than being restricted to the predefined levels L, M, Q, or H. This new `Default` value in the `ECCLevel` enum acts as an unspecified placeholder, giving users the flexibility to select an appropriate ECC level while enabling the system to validate the selection.

When generating a QR code, if the ECC level is not specified, the system uses the payload's specified ECC level. If the payload's ECC level differs from the user-specified level, the system throws an exception to handle the discrepancy. This enhancement ensures backward compatibility, allowing existing codebases to function as expected while offering users the flexibility to dynamically adjust error correction levels based on their payloads' needs.

See:
- https://github.com/codebude/QRCoder/issues/512#issuecomment-2118994971

This PR also includes bug fix #527 ; please merge it first.


codebude (approved):

codebude:
That's a nice solution. I wonder if we shouldn't just include it in the next 1.x.x release. I don't see any reason why we need to keep the change for version 2.

Just a small test case (something like generator_throws_argument_exception or similar) would be nice before merging it.

Shane32:
This particular implementation is a breaking change since I’ve changed the type. But we could add an enum value like EccLevel.Default to do the same thing, which would be backwards compatible. I’ll revise.

QRCoderApiTests/net35+net40+net50+net50-windows+netstandard20/QRCoder.approved.txt
```diff
@@ -895,6 +895,7 @@ namespace QRCoder
         public static QRCoder.QRCodeData GenerateQrCode(string plainText, QRCoder.QRCodeGenerator.ECCLevel eccLevel, bool forceUtf8 = false, bool utf8BOM = false, QRCoder.QRCodeGenerator.EciMode eciMode = 0, int requestedVersion = -1) { }
         public enum ECCLevel
         {
+            Default = -1,
             L = 0,
             M = 1,
             Q = 2,
```
Shane32, line 901: Only API change is to add a new enum entry, ensuring compatibility with code built against prior versions.

QRCoder/PayloadGenerator.cs
```diff
@@ -15,7 +15,7 @@ public static class PayloadGenerator
         public abstract class Payload
         {
             public virtual int Version { get { return -1; } }
-            public virtual QRCodeGenerator.ECCLevel EccLevel { get { return QRCodeGenerator.ECCLevel.M; } }
+            public virtual QRCodeGenerator.ECCLevel EccLevel { get { return QRCodeGenerator.ECCLevel.Default; } } // null is default
```
codebude, line 18:
```suggestion
            public virtual QRCodeGenerator.ECCLevel EccLevel { get { return QRCodeGenerator.ECCLevel.Default; } }
```
As default is -1 now, we don't need the comment any longer. (Mentioning that it is -1 isn't necessary as it's interna of the Enum.)

codebude (approved):

csturm83:
Isn't it a breaking change to throw an exception when previously didn't? Might be worth mentioning in the release notes.

codebude:
@csturm83 Thanks for pointing this out. Yes, you're right - this is somehow breaking behaviour but In my opinion nevertheless totally worth it, as cases when the exception will be raised, otherwise would lead to incorrect payloads/QR codes. (Thus I assume the change won't break nearly none implementation out in the wild, because to break it, the implementations must create in correct codes at the moment.)

I added a note on this change in the release notes: 
- https://github.com/codebude/QRCoder/wiki/Release-notes#breaking-changes

PR #527, opened 2024-05-19
Fix TrimLeadingZeros

Shane32:
Bug crashes QR code generation for text "this is a test" with ECC level M.


QRCoder/QRCodeGenerator.cs
```diff
@@ -371,7 +371,7 @@ void WriteEccLevelAndVersion()
 #endif
         private static void TrimLeadingZeros(BitArray fStrEcc, ref int index, ref int count)
         {
-            while (!fStrEcc[index])
+            while (count > 0 && !fStrEcc[index])
```
Shane32, line 374: Bug occurs when all 15 bits are zero.

PR #529, opened 2024-05-19
Add more tests

Shane32:
Adds more tests to increase confidence that performance changes will not break QR code generation.

Bug fix #527 should be merged first as many tests here would not pass without the bug fix being implemented.

PR #528, opened 2024-05-19
Reorganize GenerateQrCode for better clarity

Shane32:
Within `GenerateQrCode`, moves blocks of code into local functions to enhance readability while scoping temporary variables.

No changes were made to the operational code.  Some variables were renamed to eliminate conflicts.

PR #530, opened 2024-05-19
Refactor and optimize BlockedModules

Shane32:
This PR optimizes the speed and transient memory requirements required to note which modules are blocked and cannot be used by data.  Previously a list of blocked areas was maintained, whereas now a bitmap of the entire QR code is maintaned.  This results in much faster execution but consumes more memory.  To compensate, a `[ThreadStatic]` reusable buffer is used to store the bitmap temporarily during QR code generation.

### Requirements

This PR requires these other PRs to be merged in first, so GitHub will display a proper diff:

- #527 
- #528 
- #529 

### Before

| Method              | Mean         | Error     | StdDev    | Gen0   | Allocated |
|-------------------- |-------------:|----------:|----------:|-------:|----------:|
| CreateQRCode        |     97.11 us |  0.519 us |  0.460 us | 0.4883 |    9.3 KB |
| CreateQRCodeLong    |  1,679.91 us | 16.805 us | 14.897 us | 1.9531 |  39.24 KB |
| CreateQRCodeLongest | 14,336.49 us | 35.724 us | 31.669 us |      - |  88.35 KB |

### After

| Method              | Mean        | Error     | StdDev    | Gen0   | Allocated |
|-------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode        |    71.73 us |  0.851 us |  0.754 us | 0.3662 |   7.27 KB |
| CreateQRCodeLong    | 1,314.42 us |  9.245 us |  8.195 us |      - |  33.31 KB |
| CreateQRCodeLongest | 7,216.75 us | 29.611 us | 26.250 us |      - |  79.74 KB |


codebude:
image

@Shane32 I merged the other PRs as suggested, but it seems there's still a conflict. Can you pull the current master into your branch?

@gfoidl Another performance optimization. Would you mind doing a review?

Shane32:
@codebude No problem; done.  (No conflicts in VS, but GitHub isn't quite as smart!)

QRCoder/QRCodeGenerator.ModulePlacer.BlockedModules.cs
```diff
@@ -0,0 +1,106 @@
...
+    public partial class QRCodeGenerator
+    {
+        private static partial class ModulePlacer
+        {
+            /// <summary>
+            /// Struct that represents blocked modules using rectangles.
+            /// </summary>
+            public struct BlockedModules : IDisposable
+            {
+                private readonly bool[,] _blockedModules;
```
gfoidl, line 15: Rectangular arrays have quite bad perf compared to jagged arrays (`T[][]`), because the latter are so called "single dimensional zero indexed arrays" (sz-arrays) and have betterr support within the JIT and runtime.

To make the code easier, I'd create a matrix type that is based on the sz-array, and have an indexer that makes the work with the rows/columns easier.

QRCoder/QRCodeGenerator.ModulePlacer.BlockedModules.cs
```diff
@@ -0,0 +1,106 @@
...
+
+                [ThreadStatic]
```
gfoidl, line 17: Is parallel execution (under load) expected?
Otherwise the thread-static could be omitted, and a plain static field be used. For rent/return `Interlocked`-methods can be used than, which is faster than the thread-static access.

This is a trade-off decision if parallel execution is very likely to happen.

QRCoder/QRCodeGenerator.ModulePlacer.BlockedModules.cs
Shane32: Changed to `bool[][]`

Shane32: Also, using a `BitArray[]` is another choice.

QRCoder/QRCodeGenerator.ModulePlacer.BlockedModules.cs
Shane32: You're right, Interlocked is a better choice here.  Changed.

QRCoder/QRCodeGenerator.ModulePlacer.BlockedModules.cs
Shane32: Using `BitArray[]` is perhaps 5% slower for tiny QR codes (just 5 us), but negligible for longer QR codes.  We use `BitArray` elsewhere in the codebase so perhaps it makes sense to use it here.

gfoidl (approved):
LGTM

QRCoder/QRCodeGenerator.ModulePlacer.BlockedModules.cs
Shane32: Based on your 👍 above, I've changed to `BitArray[]`, which is more similar to the rest of the codebase.

codebude:
Thanks for your work and review @Shane32 & @gfoidl

PR #531, opened 2024-05-19
Change MaskPattern.Patterns to a list

Shane32:
This code changes `MaskPattern.Patterns` to a list instead of a dictionary to hopefully make some code easier to understand.

### Existing code

The index is stored zero-based within the QR code, and I was confused by this code here:

https://github.com/codebude/QRCoder/blob/14a83f2e1c2d946e54db72cd91f3e420630e1f71/QRCoder/QRCodeGenerator.ModulePlacer.cs#L130

followed by

https://github.com/codebude/QRCoder/blob/14a83f2e1c2d946e54db72cd91f3e420630e1f71/QRCoder/QRCodeGenerator.ModulePlacer.cs#L163

and 

https://github.com/codebude/QRCoder/blob/14a83f2e1c2d946e54db72cd91f3e420630e1f71/QRCoder/QRCodeGenerator.ModulePlacer.cs#L175-L176

which is actually correct because of 

https://github.com/codebude/QRCoder/blob/14a83f2e1c2d946e54db72cd91f3e420630e1f71/QRCoder/QRCodeGenerator.ModulePlacer.cs#L185

So since the dictionary is unnecessary, I converted it to a list.  It should be faster too but I didn't check.

### New code

Now we start with this:

https://github.com/codebude/QRCoder/blob/d48a2fb2d2b95fce2cdd35dfacdff68544e60c61/QRCoder/QRCodeGenerator.ModulePlacer.cs#L118-L120

and uses are simple:

https://github.com/codebude/QRCoder/blob/d48a2fb2d2b95fce2cdd35dfacdff68544e60c61/QRCoder/QRCodeGenerator.ModulePlacer.cs#L132

https://github.com/codebude/QRCoder/blob/d48a2fb2d2b95fce2cdd35dfacdff68544e60c61/QRCoder/QRCodeGenerator.ModulePlacer.cs#L165

https://github.com/codebude/QRCoder/blob/d48a2fb2d2b95fce2cdd35dfacdff68544e60c61/QRCoder/QRCodeGenerator.ModulePlacer.cs#L187

### DIFF
--- a/QRCoder/QRCodeGenerator.ModulePlacer.cs
+++ b/QRCoder/QRCodeGenerator.ModulePlacer.cs
@@ -168,20 +168,21 @@ namespace QRCoder
                 }
 
                 // Apply the best mask pattern to the actual QR code.
+                var selectedPatternFunc = MaskPattern.Patterns[selectedPattern.Value];
                 for (var x = 0; x < size; x++)
                 {
                     for (var y = 0; y < x; y++)
                     {
                         if (!IsBlocked(new Rectangle(x, y, 1, 1), blockedModules))
                         {
-                            qrCode.ModuleMatrix[y][x] ^= MaskPattern.Patterns[selectedPattern.Value](x, y);
-                            qrCode.ModuleMatrix[x][y] ^= MaskPattern.Patterns[selectedPattern.Value](y, x);
+                            qrCode.ModuleMatrix[y][x] ^= selectedPatternFunc(x, y);
+                            qrCode.ModuleMatrix[x][y] ^= selectedPatternFunc(y, x);
                         }
                     }
 
                     if (!IsBlocked(new Rectangle(x, x, 1, 1), blockedModules))
                     {
-                        qrCode.ModuleMatrix[x][x] ^= MaskPattern.Patterns[selectedPattern.Value](x, x);
+                        qrCode.ModuleMatrix[x][x] ^= selectedPatternFunc(x, x);
                     }
                 }
                 return selectedPattern.Value;
--- a/QRCoder/QRCodeGenerator.ModulePlacer.cs
+++ b/QRCoder/QRCodeGenerator.ModulePlacer.cs
@@ -115,8 +115,10 @@ namespace QRCoder
 
                 // Temporary QRCodeData object to test different mask patterns without altering the original.
                 var qrTemp = new QRCodeData(version);
-                foreach (var pattern in MaskPattern.Patterns)
+                for (var maskPattern = 0; maskPattern < 8; maskPattern++)
                 {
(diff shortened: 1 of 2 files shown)


Shane32:
I'll need to fix conflicts here

QRCoder/QRCodeGenerator.ModulePlacer.cs
```diff
@@ -160,29 +162,30 @@ public static int MaskCode(QRCodeData qrCode, int version, List<Rectangle> block
...
                     if (!selectedPattern.HasValue || patternScore > score)
                     {
-                        selectedPattern = pattern.Key;
+                        selectedPattern = maskPattern;
                         patternScore = score;
                     }
                 }
 
                 // Apply the best mask pattern to the actual QR code.
+                var selectedPatternFunc = MaskPattern.Patterns[selectedPattern.Value];
```
codebude, line 171: Why `selectedPattern.Value`? Shouldn't simply `selectedPattern` work, too, as selectedPattern should be an int (due to `selectedPattern = maskPattern`)?

QRCoder/QRCodeGenerator.ModulePlacer.cs
```diff
@@ -160,29 +162,30 @@ public static int MaskCode(QRCodeData qrCode, int version, List<Rectangle> block
...
                     }
 
                     if (!IsBlocked(new Rectangle(x, x, 1, 1), blockedModules))
                     {
-                        qrCode.ModuleMatrix[x][x] ^= MaskPattern.Patterns[selectedPattern.Value](x, x);
+                        qrCode.ModuleMatrix[x][x] ^= selectedPatternFunc(x, x);
                     }
                 }
-                return selectedPattern.Value - 1;
+                return selectedPattern.Value;
```
codebude, line 188: Same here. Shouldn't this be `selectedPattern`?

codebude:
Can you also have a look on the merge conflicts?
image

QRCoder/QRCodeGenerator.ModulePlacer.cs
Shane32: it's an `int?` since it's `null` initially.  it will set a value on the first pass through the loop, so it will never be null here.

Shane32: Changed to `int`, by initializing the initial score to `int.MaxValue` so all patterns score better than the variable's initial values

codebude: Sorry, my fault. I was to focused on the diff view and did not check the type of selectedPattern. Then assumed it must be int because maskPattern is of type int.

PR #534, opened 2024-05-22
Refactor compatibility extension methods

Shane32:
## Summary

Refactors the `Stream.CopyTo` method into a polyfill for .NET 3.5 only, and has `IsNullOrWhiteSpace` inline to the native implementation for available platforms.  These methods are internal and the changes will not affect operation.

In addition, `ReverseString` and `IsAllDigit` methods were removed as they were not referenced within the codebase.

## Test Plan

Existing tests cover both .NET 3.5 and other platforms, ensuring compatibility.  The changed and removed methods are not publicly accessible.

### DIFF
--- a/QRCoder/Extensions/StreamExtensions.cs
+++ b/QRCoder/Extensions/StreamExtensions.cs
@@ -1,13 +1,12 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-
-namespace QRCoder.Framework4._0Methods
+﻿#if NET35
+namespace QRCoder
 {
-    class Stream4Methods
+    internal static class StreamExtensions
     {
-        public static void CopyTo(System.IO.Stream input, System.IO.Stream output)
+        /// <summary>
+        /// Copies a stream to another stream.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


codebude (commented):
Can we rewrite/use the `IsNullOrWhiteSpace` function as an extension method in the same manner as `CopyTo`? Currently at least in the PayloadGenerator.cs the calls aren't looking nice...

https://github.com/codebude/QRCoder/blob/923dbe50310a188ff087e05c21f508bfb316f823/QRCoder/PayloadGenerator.cs#L2037

`if (Secret.IsNullOrWhiteSpace())` would look/feel smoother.

I'm sorry to mention such small things. But if we're going to tackle the points, I prefer to do it 100%. If I get too picky, let me know. Then I slow myself down a little. ;-)

Shane32:
Certainly. I’ll double check and see. I know the native method is not an extension method. So we can’t just polyfill it like CopyTo. But we could alter the name slightly if nothing else. I’ll check.

Shane32:
@codebude Looks like it works as anticipated !

codebude (approved):

marc81-marc:
Un gros merci Doc you really come from the future ☝️

PR #558, opened 2024-06-01
[WIP/QRCoder2] Fluent API experimentation PR

Shane32:
## Summary

Experimentation with fluent API syntax

## Samples

```cs
var code = QRCodeBuilder.CreateEmail("test@microsoft.com")
    .WithSubject("Testing")
    .WithBody("Hello World!")
    .WithErrorCorrection(QRCodeGenerator.ECCLevel.H)
    .RenderWith<AsciiRenderer>()
    .WithQuietZone(false)
    .ToString();

var image = QRCodeBuilder.CreatePhoneNumber("1234567890")
    .WithErrorCorrection(QRCodeGenerator.ECCLevel.H)
    .RenderWith<SystemDrawingRenderer>(20)
    .WithQuietZone(false)
    .ToBitmap();

var base64 = QRCodeBuilder.CreateMMS("1234567890", "Hello")
    .RenderWith<PngRenderer>()
    .ToBase64String();
```

## Intellisense samples

image

image

## Preliminary Findings

1. Could start with `new QRCodeBuilder.Email("test@microsoft.com")` rather than `QRCodeBuilder.CreateEmail("test@microsoft.com")` - which is pretty similar to the existing `new PayloadGenerator.Wifi()` syntax actually, with the difference being the fluent syntax for configuration.  One benefit of using `new` is that additional payloads can be added by simply adding a new class into the correct namespace.  Whereas methods cannot be added to `QRCodeBuilder` outside of QRCoder.
2. Using `RenderWith<T>` breaks the intellisense pattern because T does not give a list of the specific renderers available.  Probably better to create an extension method for each renderer, like `RenderAsAscii()` and `RenderAsPng(20)` instead
3. `RenderWith<T>` does not allow for render-specific constructor parameters.  I've compensated with support for only two patterns: (1) no arguments (2) pixels per module argument.  Having `RenderAs...()` dedicated methods would allow each renderer to require specific arguments.
4. The builder syntax could likely be added to v1.x as a new layer, as shown in this PR, with any implementation changes we wish to make within the builder methods.  Then v2 just removes (or makes internal) all the old methods.
5. The new builders, when layered on the old code, are quite easy to write.  Once the old code is removed, additional optimizations can be added to enhance trimming support.  This trimming capability may consolidate the QRCode and ArtQRCode classes.
6. All of the supporting code can be nested as deep within namespaces as desired, since intellisense always provides the correct context-sensitive methods.  Except for `RenderWith<T>`, as noted in item 2 above.  Another reason to use `RenderAsAscii()` or similar.
7. The extension methods on the renderers (`ToArray`, `ToStream`, `ToFile`, etc) are really cool -- each renderer need only implement `ToStream` and all the other methods are available via extension methods.  Similar functionality for text renderers;  even with only implementing `ToString` they can also provide `ToFile`, `ToStream`, `ToBase64`, and so on via extension methods.

### DIFF
--- a/QRCoder/Builders/Payloads/Implementations/EmailPayload.cs
+++ b/QRCoder/Builders/Payloads/Implementations/EmailPayload.cs
@@ -1,0 +1,41 @@
+﻿using System;
+
+namespace QRCoder.Builders.Payloads.Implementations
+{
+    public class EmailPayload : PayloadBase
+    {
+        public EmailPayload(string address)
+        {
+            _address = address;
+        }
+
+        private string _address { get; set; }
+        private string _subject { get; set; }
+        private string _body { get; set; }
+        private PayloadGenerator.Mail.MailEncoding _encoding { get; set; } = PayloadGenerator.Mail.MailEncoding.MAILTO;
+
+        public EmailPayload WithSubject(string subject)
+        {
+            _subject = subject;
+            return this;
+        }
+
+        public EmailPayload WithBody(string body)
+        {
+            _body = body;
+            return this;
+        }
+
+        public EmailPayload WithEncoding(PayloadGenerator.Mail.MailEncoding encoding)
+        {
+            if (encoding != PayloadGenerator.Mail.MailEncoding.MAILTO && encoding != PayloadGenerator.Mail.MailEncoding.SMTP && encoding != PayloadGenerator.Mail.MailEncoding.MATMSG)
+            {
+                throw new ArgumentOutOfRangeException(nameof(encoding));
+            }
+            _encoding = encoding;
+            return this;
+        }
+
+        protected override string Value => new PayloadGenerator.Mail(_address, _subject, _body, _encoding).ToString();
+    }
+}
--- a/QRCoder/Builders/Payloads/Implementations/StringPayload.cs
+++ b/QRCoder/Builders/Payloads/Implementations/StringPayload.cs
@@ -1,0 +1,14 @@
+﻿namespace QRCoder.Builders.Payloads.Implementations
+{
+    public class StringPayload : PayloadBase
+    {
+        private string _data;
+
+        public StringPayload(string data)
+        {
+            _data = data;
+        }
+
+        protected override string Value => _data;
+    }
+}
--- a/QRCoder/Builders/Payloads/PayloadBase.cs
+++ b/QRCoder/Builders/Payloads/PayloadBase.cs
@@ -1,0 +1,21 @@
+﻿namespace QRCoder.Builders.Payloads
+{
+    public abstract class PayloadBase : IPayload, IConfigurableEccLevel, IConfigurableEciMode, IConfigurableVersion
+    {
+        protected virtual QRCodeGenerator.ECCLevel EccLevel { get; set; } = QRCodeGenerator.ECCLevel.Default;
+        QRCodeGenerator.ECCLevel IConfigurableEccLevel.EccLevel { get => EccLevel; set => EccLevel = value; }
+
+        protected virtual QRCodeGenerator.EciMode EciMode { get; set; } = QRCodeGenerator.EciMode.Default;
+        QRCodeGenerator.EciMode IConfigurableEciMode.EciMode { get => EciMode; set => EciMode = value; }
+
+        protected virtual int Version { get; set; } = -1;
+        int IConfigurableVersion.Version { get => Version; set => Version = value; }
+
+        protected abstract string Value { get; }
+
+        public virtual QRCodeData ToMatrix()
+        {
+            return QRCodeGenerator.GenerateQrCode(Value, EccLevel, false, false, EciMode, Version);
+        }
+    }
+}
--- a/QRCoder/Builders/Payloads/PayloadExtensions.cs
+++ b/QRCoder/Builders/Payloads/PayloadExtensions.cs
@@ -1,0 +1,47 @@
+﻿using QRCoder.Builders.Payloads;
+using QRCoder.Builders.Renderers;
+using QRCoder.Builders.Renderers.Implementations;
+
+namespace QRCoder
+{
+    public static class PayloadExtensions
+    {
+        public static T WithErrorCorrection<T>(this T payload, QRCodeGenerator.ECCLevel eccLevel)
+            where T : IConfigurableEccLevel
+        {
+            payload.EccLevel = eccLevel;
+            return payload;
+        }
+
+        public static T WithEciMode<T>(this T payload, QRCodeGenerator.EciMode eciMode)
+            where T : IConfigurableEciMode
+        {
+            payload.EciMode = eciMode;
+            return payload;
+        }
+
+        public static T WithVersion<T>(this T payload, int version)
+            where T : IConfigurableVersion
+        {
+            payload.Version = version;
+            return payload;
+        }
+
+        public static T RenderWith<T>(this IPayload payload)
+            where T : IRenderer, new()
+        {
+            var renderer = new T();
+            renderer.Payload = payload;
+            return renderer;
+        }
+
+        public static T RenderWith<T>(this IPayload payload, int pixelsPerModule)
+            where T : IRenderer, IConfigurablePixelsPerModule, new()
+        {
+            var renderer = new T();
+            renderer.Payload = payload;
+            renderer.PixelsPerModule = pixelsPerModule;
+            return renderer;
+        }
+    }
+}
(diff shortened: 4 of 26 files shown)


QRCoder/Builders/Payloads/Implementations/EmailPayload.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+
+namespace QRCoder.Builders.Payloads.Implementations
+{
+    public class EmailPayload : PayloadBase
```
Shane32, line 5: Sample where the ctor only has the email address, and the subject/body are provided by extension methods.  Realistically, I'm thinking that the required or common parameters always appear in the constructor (or as an overload), but all of the optional/rare options (like `MailEncoding`) are extension methods.

QRCoder/Builders/Payloads/Implementations/StringPayload.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace QRCoder.Builders.Payloads.Implementations
+{
+    public class StringPayload : PayloadBase
```
Shane32, line 3: We can use this class for any payload that immediately serializes to a string value.  At least if we design it like `CreateUrl(...)`.  Of course it would be nearly as easy to have additional classes for each payload type, similar to how it is now, if we want `new QRCodeBuilder.Url(...)` or whatnot.

QRCoder/Builders/Payloads/PayloadBase.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace QRCoder.Builders.Payloads
+{
+    public abstract class PayloadBase : IPayload, IConfigurableEccLevel, IConfigurableEciMode, IConfigurableVersion
```
Shane32, line 3: The idea here is that while most all payloads can inherit from `PayloadBase` and inherit some configurable settings, perhaps some payloads, like the swiss banking payload, does not support configuring the Eci mode or whatever.  Or, users can override these properties to add validation, ensuring that a certain minimum ECC level is maintained.  There's a number of ways to actually implement this, all with similar end results.

QRCoder/Builders/Payloads/PayloadBase.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        protected virtual QRCodeGenerator.ECCLevel EccLevel { get; set; } = QRCodeGenerator.ECCLevel.Default;
+        QRCodeGenerator.ECCLevel IConfigurableEccLevel.EccLevel { get => EccLevel; set => EccLevel = value; }
```
Shane32, line 6: These interfaces are always explicitly implemented so that they do not appear on the intellisense list of accessible properties for the class.  Rather the extension method will appear, which uses generics to return the same type.

If that's confusing, for example, this code would break the fluent syntax:

```cs
public PayloadBase WithEciMode(EciMode value)
{
    _eciMode = value;
}
```

Because it returns the base class type, rather than the type of the derived type.  Extension methods allow the intended behavior.

QRCoder/Builders/Payloads/PayloadExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+            return payload;
+        }
+
+        public static T RenderWith<T>(this IPayload payload)
+            where T : IRenderer, new()
+        {
+            var renderer = new T();
+            renderer.Payload = payload;
+            return renderer;
+        }
```
Shane32, line 36: This functions, but I think `RenderWithAsciiRenderer()` or `RenderAsPng()` would be better than generics here.  I specifically don't like that intellisense does not show the compatible renderers for `T`.  Sorta breaks the goal.

image

QRCoder/Builders/Renderers/IStreamRenderer.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.IO;
+
+namespace QRCoder.Builders.Renderers
+{
+    public interface IStreamRenderer
+    {
+        MemoryStream ToStream();
+    }
```
Shane32, line 8: Concept is that all byte-array-type renderers (or ones that can serialize to a byte array) implement this interface.  If they use a `MemoryStream` under the hood, it's a direct pass-through.  If they generate a byte array directly, then wrapping it in a `MemoryStream` is quite inexpensive, as it does not duplicate the entire byte array.

QRCoder/Builders/Renderers/RendererExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        public static byte[] ToArray(this IStreamRenderer streamRenderer)
+        {
+            var memoryStream = streamRenderer.ToStream();
+#if NETSTANDARD || NETCOREAPP // todo: target .NET Framework 4.6 or newer so this code path is supported
+            // by using TryGetBuffer, there is extremely small consequence to wrapping a byte[] in a MemoryStream temporarily
+            if (memoryStream.TryGetBuffer(out var buffer) && buffer.Count == buffer.Array.Length)
+            {
+                return buffer.Array;
+            }
```
Shane32, line 38: And here we can extract the original array without duplicating it.

QRCoder/Builders/Renderers/ITextRenderer.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.IO;
+
+namespace QRCoder.Builders.Renderers
+{
+    public interface ITextRenderer
+    {
+        string ToString();
+    }
```
Shane32, line 8: Concept is that all string-based renderers (Ascii, Svg) implement this interface.

QRCoder/Builders/Renderers/Implementations/SystemDrawingRenderer.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public Bitmap ToBitmap()
+        {
+            return new QRCode(QrCodeData).GetGraphic(
+                _pixelsPerModule, _darkColor, _lightColor, 
+                _icon, (int)Math.Round(_iconSizePercent), _iconBorderWidth, 
+                QuietZone, _iconBackgroundColor);
+        }
+
+        public MemoryStream ToStream()
```
Shane32, line 52: Renderers which return a specialized type may optionally also implement `IStreamRenderer` as is shown here.  Then `ToFile` and other extension methods for byte-array type renderers are available.

QRCoder/Builders/Renderers/RendererBase.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace QRCoder.Builders.Renderers
+{
+    public abstract class RendererBase : IRenderer, IConfigurableQuietZones
+    {
+        protected QRCodeData QrCodeData { private set; get; }
+
+        protected bool QuietZone { get; set; } = true;
+        bool IConfigurableQuietZones.QuietZone { get => QuietZone; set => QuietZone = value; }
+        IPayload IRenderer.Payload { set => QrCodeData = value.ToMatrix(); }
```
Shane32, line 11: Another option would be to store the `IPayload` object itself.  I opted to call `ToMatrix` immediately, so if there were any exceptions while converting the payload to a module matrix, it would be caught at this point, enhancing the ability to diagnose faulty fluent syntax.

QRCoder/Builders/Renderers/RendererExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+            return new ArraySegment<byte>(memoryStream.ToArray());
+        }
+
+        public static string ToBase64String(this IStreamRenderer streamRenderer)
+        {
+            var data = ToArraySegment(streamRenderer);
+            return Convert.ToBase64String(data.Array, data.Offset, data.Count);
+        }
+
+        public static void ToFile(this IStreamRenderer streamRenderer, string fileName)
```
Shane32, line 76: Note that if any renderer wanted to optimize something like `ToFile`, they could implement the method directly, which takes precedence over extension methods.

csturm83:
> The builder syntax could likely be added to v1.x **as a new layer**, as shown in this PR

`QRCoder2.FluentExtensions` maybe? 😉

codebude:
Hi Shane, thanks for the PR. That looks pretty interesting. However, I don't want to merge this into v1.x, but would like to include it in v2 first, as I don't want to make/document/communicate any more major API changes to v1. I hope that's okay with you.

Shane32:
Certainly. I wrote this more of a discussion/preview/review of a potential approach.

codebude:
Hi @Shane32 ,  

codebude:
Closed in consultation with @Shane32

PR #559, opened 2024-06-01
Add NRT metadata

Shane32:
## Summary

Adds nullable reference type (NRT) metadata to all members

## Notes

This upgrades the C# version enough to support NRT.  NRT is not incompatible with older frameworks, and no new features are being used that would cause an issue.

## Test plan
Existing tests demonstrate compatibility.  Existing API tests demonstrate new NRT metadata exposed with no source/binary breaking API changes.


QRCoder/PayloadGenerator/Mail.cs
```diff
@@ -19,7 +19,7 @@ public class Mail : Payload
             /// <param name="subject">Subject line of the email</param>
             /// <param name="message">Message content of the email</param>
             /// <param name="encoding">Payload encoding type. Choose dependent on your QR Code scanner app.</param>
-            public Mail(string mailReceiver = null, string subject = null, string message = null, MailEncoding encoding = MailEncoding.MAILTO)
+            public Mail(string? mailReceiver = null, string? subject = null, string? message = null, MailEncoding encoding = MailEncoding.MAILTO)
```
Shane32, line 22: The fact that `mailReceiver` is optional is pretty dumb.  There's no other way to set the target email address, and a QR code for an email address that has no email address seems dumb to me.  But I didn't change it for this PR.  Note that we could make this property non-optional, which would still be binary-compatible but not source-compatible.  I'd recommend that change within 1.x.

QRCoder/ArtQRCode.cs
```diff
@@ -211,8 +211,6 @@ private bool IsPartOfFinderPattern(int x, int y, int numModules, int offset)
         /// <returns>Resized image as bitmap</returns>
         private Bitmap Resize(Bitmap image, int newSize)
         {
-            if (image == null) return null;
```
Shane32, line 214: already checked by the caller

QRCoder/QRCodeGenerator/Polynom.cs
```diff
@@ -258,19 +258,14 @@ void ThrowArgumentOutOfRangeException()
             private static void ReturnArray(PolynomItem[] array)
             {
                 if (array == null)
-                    ThrowArgumentNullException();
```
Shane32, line 261: Dispose calls this private method and Dispose should not throw when called twice

QRCoder/QRCoder.csproj
```diff
@@ -9,6 +9,8 @@
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
 	<IsTrimmable Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</IsTrimmable>
+	<Nullable>enable</Nullable>
+	<LangVersion>9.0</LangVersion>
```
Shane32, line 13: In the future we may want to bump this higher . For now I only bumped as necessary for this PR.

QRCoder/PayloadGenerator/Mail.cs
codebude: Making the receiver nullable was a conscious decision. What happens when someone scans the mail QR code? The email app is opened and pre-populated with the values from the QR code. For things like "give us feedback" emails, I understand that there should always be a recipient. 

However, there are also use cases where it makes no sense to pre-populate the recipient. Example: A store wants to offer customers the opportunity to recommend their business to friends of the customer. They create a QR code with a text like "Hey, I'm in store XYZ right now. You should go there too". The QR code is displayed in the shop window, etc. A customer scans it, receives the preformatted recommendation mail on their phone and only has to type in the email address of the friend they want to send the recommendation to.

TL;DR: No recipient therefore always makes sense if you want to use the QR codes as templates for emails.

Shane32: Nice. I hadn’t thought of that.

codebude (approved):

PR #560, opened 2024-06-02
Remove framework overrides

Shane32:
## Summary

The QRCoder solution won't build on my Windows 11 laptop with a fresh installation of Visual Studio 2022, even with the .NET Framework 3.5 development tools installed.  It seems to be related to the framework override csproj properties.  It builds fine when I remove them.

Can we remove these properties?

## Test plan
Ensure it builds without these excess properties, both locally and in CI.

### DIFF
--- a/QRCoder.Xaml/QRCoder.Xaml.csproj
+++ b/QRCoder.Xaml/QRCoder.Xaml.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>net35;net40;net5.0-windows;net6.0-windows</TargetFrameworks>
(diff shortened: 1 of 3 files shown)


codebude:
The framework overrides were added a good 4 years ago by @csturm83 via https://github.com/codebude/QRCoder/commit/0d084052bc8957fa12766deddc465903afe8c4f3. Maybe he can still remember why. 

Unfortunately, I no longer know. I can only assume that it was due to the build pipeline at the time via MyGet, where the .NET framework was not in the default path. But I'm not 100 percent sure. If csturm83 doesn't veto it, we can merge it.

Shane32:
How do we test if it is needed anymore? Dotnet pack from within CI?

csturm83:
@Shane32 Not sure. There was an update in the comment that was referenced in the workaround (https://github.com/dotnet/msbuild/issues/1333#issuecomment-296346352) to reference ref assemblies via a [NuGet package](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/). You could try that if removing the overrides doesn't work.

codebude:
> You could try that if removing the overrides doesn't work.

I would say it works. At least the pipeline logs say that the net35 binaries were successfully build: https://github.com/codebude/QRCoder/actions/runs/9334333660/job/25692441349?pr=560#step:5:13

Shane32:
I had problems a while back where I was required to add the NuGet package reference to the .NET Reference Assemblies.  However, I don't need to anymore, and in fact it breaks builds within the latest .NET SDK.  See link below.  The .NET team recommended that I remove the reference, even though there still exists MS documentation stating to include the reference for certain scenarios.  Supposedly the reference is automatically included by the .NET SDK for .NET Framework builds or something.  I would assume that the issue requiring the patch has since been resolved.

See: https://github.com/dotnet/aspnetcore/issues/54544

csturm83:
> the net35 binaries were successfully _**build**_:

The original issue was surrounding `dotnet pack`. I suspect the issue in the SDK has since been resolved as @Shane32 suggests.

codebude (approved):

PR #561, opened 2024-06-02
Add XML comments to all public members

Shane32:
## Summary

This PR adds XML comments to all public members

No code has been changed.

## Test Plan

CS1591 is enforced so any public members added in the future will require XML comments or will break the build.

## Notes

These xml comments may not be perfect, but they're better than nothing.  Please suggest any changes desired.


QRCoder/PayloadGenerator/Girocode.cs
```diff
@@ -49,22 +54,26 @@ public Girocode(string iban, string bic, string name, decimal amount, string rem
                 if (amount.ToString().Replace(",", ".").Contains(".") && amount.ToString().Replace(",",".").Split('.')[1].TrimEnd('0').Length > 2)
                     throw new GirocodeException("Amount must have less than 3 digits after decimal point.");
                 if (amount < 0.01m || amount > 999999999.99m)
-                    throw new GirocodeException("Amount has to at least 0.01 and must be smaller or equal to 999999999.99.");
+                    throw new GirocodeException("Amount has to be at least 0.01 and must be smaller or equal to 999999999.99.");
```
codebude, line 57: Can you please fix the test cases that do assertions on this exception? (See Github action runner. Test cases seem to be broken.)

codebude (changes requested):
Wow, that looks like a lot of work. Thanks! Before merging I kindly ask you to fix the test cases, too.

QRCoder/PayloadGenerator/Girocode.cs
Shane32: Certainly

codebude (approved):

PR #562, opened 2024-06-05
Reformat code

Shane32:
## Summary

This PR adds an `.editorconfig` file along with a workflow to verify that the code matches the workflow.  Code throughout the solution has been reformatted to match.  Nearly all build warnings are configured as errors, enforcing proper code style both in the editor and via a test workflow.

## Process

In general, each warning that I was able to auto-fix, I made a separate commit.  Then a few commits to get the CI working and a few more to clean up some last few warnings.

## Test plan

- The code tests and API tests ensure that the code functions correctly and no public API changes have been made.
- A new workflow has been added to verify that the formatting matches the configured requirements.  This does not support the UWP demo app, probably because it's not a SDK format csproj.  Also it does not support the .NET Standard 1.3 or .NET Core 1.x TFMs.

## Review

To reduce the diff in GitHub, hide whitespace:

image


Shane32:
@codebude Ready for review whenever you're ready; no rush.

codebude (approved):

PR #565, opened 2024-06-19
Updated Bitmap header

Apflkuacha:
## Summary
Updated the Bitmap header from BITMAPCOREHEADER to BITMAPINFOHEADER according to [https://en.wikipedia.org/wiki/BMP_file_format#DIB_header_(bitmap_information_header)](https://en.wikipedia.org/wiki/BMP_file_format#DIB_header_(bitmap_information_header))

Slightly newer version, now with bitmap size in header.

### What existing problem does the pull request solve?**
Updating the header was necessary as inserting the previous version of the BMP wasn't possible into a PDF document.


QRCoder/BitmapByteQRCode.cs
```diff
@@ -97,9 +98,12 @@ public byte[] GetGraphic(int pixelsPerModule, byte[] darkColorRgb, byte[] lightC
                 }
             }
 
-            //finalize with terminator
-            bmp.AddRange(new byte[] { 0x00, 0x00 });
```
Shane32, line 101: Terminator removed?

Apflkuacha: Yeah i removed it, I haven't found any information that a terminator is necessary in the bitmap file. The necessary zeros that it is 4 byte aligned are already added in line 90 to 96: `if (sideLength%4 != 0) ...`

Apflkuacha: Just double-checked once again with the same QRcode.bmp saved from Paint, the result: the file is the same except two bytes are different: `the horizontal and vertical resolution of the image. (pixel per metre, signed integer)`. But no terminator bytes are created by other apps which create .bmp files.

Shane32: > Just double-checked once again with the same QRcode.bmp saved from Paint, the result: the file is the same except two bytes are different: `the horizontal and vertical resolution of the image. (pixel per metre, signed integer)`. That depends on the use case. But no terminator bytes are created by other apps which create .bmp files.

Perfect!

codebude: For the life of me, I can't remember why I wrote the terminator. I certainly didn't make it up, but I can't find any source that mentions this type of terminator. So as far as I'm concerned, it can go.

QRCoder/BitmapByteQRCode.cs
```diff
@@ -63,7 +63,7 @@ public byte[] GetGraphic(int pixelsPerModule, byte[] darkColorRgb, byte[] lightC
             List<byte> bmp = new List<byte>();
 
             //header
-            bmp.AddRange(new byte[] { 0x42, 0x4D, 0x4C, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x1A, 0x00, 0x00, 0x00, 0x0C, 0x00, 0x00, 0x00 });
```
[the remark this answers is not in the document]
Apflkuacha: feel free to change it, the file size is now set at the end of the method starting from line 102: `var bmpFileSize = IntTo4Byte(bmp.Count); ...`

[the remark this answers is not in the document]
codebude: Looks good to me. Thanks!

codebude (approved):

codebude:
Merged and updated the release notes: https://github.com/codebude/QRCoder/wiki/Release-Notes/_compare/f65174dfcfa139487de893bd439bbd5b659ca48e...9279d42fbdb335d5f8cdb907b7fc077895093ea4

Shane32:
@Apflkuacha can I ask, how exactly were you attempting to insert the bmp into a PDF?  With Acrobat?  With itextsharp?

Apflkuacha:
Yeah, that I mentioned withoud details. I was using PDFsharp to insert that bmp and with the old header it couldn't identify the image and with the newer header it works.

PR #569, opened 2024-06-29
Split PayloadGeneratorTests into multiple files

Shane32:
## Summary
Test cleanup.  Splits PayloadGeneratorTests into multiple files just like how the generator code was broken into multiple files.  No changes to tests.


codebude (approved):

codebude:
Looks good to me. Thanks!

codebude:
@Shane32 do you plan any further PRs that should be part of v1.6? Otherwise I would release 1.6 this weekend.

Shane32:
No. Sounds great; go ahead.

Shane32:
Unless you want me to write a pr that changes the bmp writer to make 1-bit images with indexed color like the png one does instead of 24-bit uncompressed images. The generated image files would be much smaller.

codebude:
That sounds like a great idea. But we can also implement it in a 1.6.1/1.7.0 (or if nothing else is on the roadmap) in a 2.0. ;-)
So no need to rush. :-)

PR #591, opened 2025-04-09
Pin .NET SDK to 8.x for local/CI builds

Shane32:
Ensures that test runs are consistent whether or not the CI runner or local devs have the same SDK(s) installed


gfoidl:
The change to `QRCoder.sln` is missing here.
https://github.com/codebude/QRCoder/blob/5a97eb2ba27aa065405b935840d7a6636355554f/QRCoder.sln#L28

Shane32:
Thanks @gfoidl .  That's what I get when I try to edit from GitHub.com !

codebude (approved):

PR #592, opened 2025-04-13
Add Micro QR code support

Shane32:
Prerequisites (to reduce diff):
- #590 

Status:
- All generated micro QR codes are scannable
- Most QR codes match the output of the segno python library
- Need to add tests and validate combinations of lengths, character types, etc

Side notes:
- ASCII small generator has x/y swapped and prints inverted compared to regular ASCII generator
- The segno python library does not apply padding properly so micro QR codes consisting mostly of padding don't match

PR #595, opened 2025-06-16
Reduced some allocations in `QRCodeGenerator` (NETCORE_APP only)

gfoidl:
Profiling showed that there are some allocations in `QRCodeGenerator` that can be quite easily avoided.

<details>
  <summary>simple console app for profiling</summary>

```c#
using System.Diagnostics;
using QRCoder;

string payload = $"""
    Time    : {DateTimeOffset.Now:O}
    Machine : {Environment.MachineName}
    Activity: {Activity.Current?.RootId}
    """;

int len = 0;

for (int i = 0; i < 1_000; ++i)
{
    string imgSrc = GetQRCode("sdfsdf");
    len += imgSrc.Length;
}

Console.WriteLine(len);

static string GetQRCode(string payload)
{
    using QRCodeGenerator qrCodeGenerator = new();
    using QRCodeData qrCodeData = qrCodeGenerator.CreateQrCode(payload, QRCodeGenerator.ECCLevel.Q);
    using Base64QRCode qrCodeBase64 = new(qrCodeData);
    string base64QRCode = qrCodeBase64.GetGraphic(20);
    return $"data:image/png;base64,{base64QRCode}";
}
```
</details>

Allocations are removed / avoided for:
* `Dictionary<,>.Entry[]`
* `QRCodeGenerator.PolynomItem[]`
* `Dictionary<,>`

The change is done only for .NET (Core) targets, as `Span<T>` is used.
By adding a reference to System.Memory package this change could also be done for .NET Desktop.

### Before

before

### After

after

### Before

```
| Method              | Mean        | Error     | StdDev    | Gen0   | Allocated |
|-------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode        |    235.2 μs |   4.50 μs |   5.00 μs | 2.1973 |    7.2 KB |
| CreateQRCodeLong    |  2,684.9 μs |  48.00 μs |  44.89 μs | 7.8125 |  33.25 KB |
| CreateQRCodeLongest | 16,264.4 μs | 317.90 μs | 485.47 μs |      - |  79.69 KB |
```

### After

```
| Method              | Mean        | Error     | StdDev    | Gen0   | Allocated |
|-------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode        |    232.0 μs |   4.63 μs |   6.64 μs | 0.9766 |   4.38 KB |
| CreateQRCodeLong    |  2,574.1 μs |  41.97 μs |  39.26 μs | 3.9063 |     12 KB |
| CreateQRCodeLongest | 15,573.7 μs | 292.90 μs | 259.65 μs |      - |  48.24 KB |
```

### DIFF
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1061,6 +1061,17 @@ public partial class QRCodeGenerator : IDisposable
 #if NETCOREAPP
         static ReadOnlySpan<int> GetNotUniqueExponents(Polynom list, Span<int> buffer)
         {
+            // It works as follows:
+            // 1. a scratch buffer of the same size as the list is passed in
+            // 2. exponents are written / copied to that scratch buffer
+            // 3. scratch buffer is sorted, thus the exponents are in order
+            // 4. for each item in the scratch buffer (= ordered exponents) it's compared w/ the previous one
+            //   * if equal, then increment a counter
+            //   * else check if the counter is $>0$ and if so write the exponent to the result
+            // 
+            // For writing the result the same scratch buffer is used, as by definition the index to write the result 
+            // is `<=` the iteration index, so no overlap, etc. can occur.
+
             Debug.Assert(list.Count == buffer.Length);
 
             int idx = 0;
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1018,7 +1018,13 @@ public partial class QRCodeGenerator : IDisposable
 
         // Identify and merge terms with the same exponent.
         var toGlue = GetNotUniqueExponents(resultPolynom);
+#if NETCOREAPP
+        var gluedPolynoms = toGlue.Length <= 128
+            ? stackalloc PolynomItem[128].Slice(0, toGlue.Length)
+            : new PolynomItem[toGlue.Length];
+#else
         var gluedPolynoms = new PolynomItem[toGlue.Length];
+#endif
         var gluedPolynomsIndex = 0;
         foreach (var exponent in toGlue)
         {
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1042,7 +1042,7 @@ public partial class QRCodeGenerator : IDisposable
 
         // Remove duplicated exponents and add the corrected ones back.
         for (int i = resultPolynom.Count - 1; i >= 0; i--)
-            if (toGlue.Contains(resultPolynom[i].Exponent))
+            if (Array.IndexOf(toGlue, resultPolynom[i].Exponent) >= 0)
                 resultPolynom.RemoveAt(i);
         foreach (var polynom in gluedPolynoms)
             resultPolynom.Add(polynom);
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -224,15 +224,9 @@ public partial class QRCodeGenerator : IDisposable
         // Place interleaved data on module matrix
         var qrData = PlaceModules();
 
-#if NETCOREAPP
-        foreach (var codeWord in codeWordWithECC)
-        {
-            codeWord.Dispose();
-        }
-#endif
+        CodewordBlock.ReturnList(codeWordWithECC);
 
         return qrData;
-
 
         // fills the bit array with a repeating pattern to reach the required length
         void PadData()
@@ -268,7 +262,7 @@ public partial class QRCodeGenerator : IDisposable
             using (var generatorPolynom = CalculateGeneratorPolynom(eccInfo.ECCPerBlock))
             {
                 //Calculate error correction words
-                codewordBlocks = new List<CodewordBlock>(eccInfo.BlocksInGroup1 + eccInfo.BlocksInGroup2);
+                codewordBlocks = CodewordBlock.GetList(eccInfo.BlocksInGroup1 + eccInfo.BlocksInGroup2);
                 AddCodeWordBlocks(1, eccInfo.BlocksInGroup1, eccInfo.CodewordsInGroup1, 0, bitArray.Length, generatorPolynom);
                 int offset = eccInfo.BlocksInGroup1 * eccInfo.CodewordsInGroup1 * 8;
                 AddCodeWordBlocks(2, eccInfo.BlocksInGroup2, eccInfo.CodewordsInGroup2, offset, bitArray.Length - offset, generatorPolynom);
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -218,6 +218,13 @@ public partial class QRCodeGenerator : IDisposable
         // Place interleaved data on module matrix
         var qrData = PlaceModules();
 
+#if NETCOREAPP
+        foreach (var codeWord in codeWordWithECC)
+        {
+            codeWord.Dispose();
+        }
+#endif
+
         return qrData;
 
 
@@ -287,7 +294,7 @@ public partial class QRCodeGenerator : IDisposable
             for (var i = 0; i < eccInfo.ECCPerBlock; i++)
             {
                 foreach (var codeBlock in codeWordWithECC)
-                    if (codeBlock.ECCWords.Length > i)
+                    if (codeBlock.ECCWords.Count > i)
                         length += 8;
             }
             length += CapacityTables.GetRemainderBits(version);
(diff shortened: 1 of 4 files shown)


QRCoder/QRCodeGenerator.cs
```diff
@@ -1017,8 +1018,15 @@ private static Polynom MultiplyAlphaPolynoms(Polynom polynomBase, Polynom polyno
         }
 
         // Identify and merge terms with the same exponent.
+#if NETCOREAPP
+        var toGlue = GetNotUniqueExponents(resultPolynom, resultPolynom.Count <= 128 ? stackalloc int[128].Slice(0, resultPolynom.Count) : new int[resultPolynom.Count]);
```
gfoidl, line 1254: Unfortunately I don't know too little about the QR-specs, but the `resultPolynom.Count <= 128` could be avoided if by spec the count can't be as high. Or we can change the threshold to a higher value.

If the count can't be as high, then the fallback to the array allocation could also be removed.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1036,7 +1044,11 @@ private static Polynom MultiplyAlphaPolynoms(Polynom polynomBase, Polynom polyno
 
         // Remove duplicated exponents and add the corrected ones back.
         for (int i = resultPolynom.Count - 1; i >= 0; i--)
+#if NETCOREAPP
             if (toGlue.Contains(resultPolynom[i].Exponent))
+#else
+            if (Array.IndexOf(toGlue, resultPolynom[i].Exponent) >= 0)
```
gfoidl, line 1282: Before Linq was used w/ it's generic `Contains` method.
* for AOT Linq requires way more code
* it's an interface dispatch that isn't needed

Note: for .NET (Core) the span-based `Contains` is used.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1046,20 +1058,55 @@ private static Polynom MultiplyAlphaPolynoms(Polynom polynomBase, Polynom polyno
         return resultPolynom;
 
         // Auxiliary function to identify exponents that appear more than once in the polynomial.
-        int[] GetNotUniqueExponents(Polynom list)
+#if NETCOREAPP
+        static ReadOnlySpan<int> GetNotUniqueExponents(Polynom list, Span<int> buffer)
```
gfoidl, line 1294: This implementation doesn't need a `Dictionary<,>` to determine the non unique exponents.

It works as follows:
1. a scratch buffer of the same size as the list is passed in
2. exponents are written / copied to that scratch buffer
3. scratch buffer is sorted, thus the exponents are in order
4. for each item in the scratch buffer (= ordered exponents) it's compared w/ the previous one
  * if equal, then increment a counter
  * else check if the counter is $>0$ and if so write the exponent to the result

For writing the result the same scratch buffer is used, as by definition the index to write the result is `<=` the iteration index, so no overlap, etc. can occur.

That way we avoid the need for a second scratch buffer.

---

Should someting like this be added as comment or is it cleare enough how it works?

Shane32: The function isn't long; so long as the method comment is descriptive enough, I think it's fine.

QRCoder/QRCodeGenerator.cs
Shane32: When running the tests, it doesn't go over 64.  Seems fine to me.

QRCoder/QRCodeGenerator.cs
```diff
@@ -1046,20 +1058,55 @@ private static Polynom MultiplyAlphaPolynoms(Polynom polynomBase, Polynom polyno
...
         {
-            var dic = new Dictionary<int, bool>(list.Count);
+            Debug.Assert(list.Count == buffer.Length);
```
Shane32, line 1307: Do you plan to leave the `Debug.Assert` code in here?  Just wondering; it will get removed from release builds anyway.

Shane32:
> By adding a reference to System.Memory package this change could also be done for .NET Desktop.

I wouldn't.  There has been so many performance enhancements in .NET Core since .NET Framework, that if anyone wants better performance, they should use .NET Core.  And I'm sure the fallback performance is plenty good enough anyway.

Shane32:
The reduction in allocations is very impressive!

Shane32:
I fixed the CI scripts in #592 btw

Shane32 (approved):
Looks good to me! (subject to tests passing, of course)

QRCoder/QRCodeGenerator.cs
gfoidl: I like to have some `Debug.Assert`s 
* they check some invariants in debug builds and while running tests
* are a bit of self-documenting the intention (so why write a comment, when the assert can also be done?)

Especially here the `buffer` is passed in as argument and must have the correct size.
If the size doesn't match, then the tests (under debug) will fail and one knows why. Otherwise it may be hard to track down the bug.

So I'd leave them in the code.

As you said: for !DEBUG these asserts won't have any effect.

QRCoder/QRCodeGenerator.cs
gfoidl: My tests showed the same (what a wonder 😉).
But in that regard I'm a bit paranoid...in the sense wenn the fallback is removed what will happen when a bigger buffer is needed (maybe now by some special inputs or anytime in the future when new additions may be done)? This would result in an exception. W/ the fallback there's a safety net.

Except it can be proven via the QR-spec that the count never will exceed a certain threshold (but does that hold in the future too?)

If the non-stackalloc path would be very frequent, then renting from the array-pool would be an option, but here it's a assumed to be rare / never taken path.

Would you still remove the fallback or let's just leave it as is?

gfoidl:
> I fixed the CI scripts in #592 btw

Should that change be separated into a own PR to fix CI?

Otherwise every PR has to re-do the same until your PR gets merged.

-------

@Shane32 thanks for your review and comments 👍🏻

QRCoder/QRCodeGenerator.cs
Shane32: Oh I’d definitely leave the fallback. There’s just no reason to remove it. The JIT will realize it’s not a common pathway and optimize it accordingly. But as you say, we neither of us know enough about how these polynomials work to know if it will always be < 128.

Shane32:
> Should that change be separated into a own PR to fix CI?

Honestly yes…

gfoidl:
Found some more allocation that can be quite easily avoided.

```
| Method              | Mean        | Error     | StdDev    | Gen0   | Allocated |
|-------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode        |    240.0 μs |   4.57 μs |   4.89 μs | 1.2207 |   4.07 KB |
| CreateQRCodeLong    |  2,730.3 μs |  41.85 μs |  37.10 μs |      - |  11.05 KB |
| CreateQRCodeLongest | 17,367.9 μs | 345.33 μs | 657.03 μs |      - |  46.42 KB |
```

(note: time column isn't really comparable to results in top post (different machine), so only allocations count here)

QRCoder/QRCodeGenerator/CapacityTables.cs
```diff
@@ -36,7 +37,17 @@ private static class CapacityTables
         /// block group details, and other parameters required for encoding error correction data.
         /// </returns>
         public static ECCInfo GetEccInfo(int version, ECCLevel eccLevel)
-            => _capacityECCTable.Single(x => x.Version == version && x.ErrorCorrectionLevel == eccLevel);
```
gfoidl, line 40: The downside of Linq is that no (generic) state can be passed in, thus a closure is needed, and this one showed up in a memory-profile.

QRCoder/QRCodeGenerator/CapacityTables.cs
```diff
@@ -92,11 +103,19 @@ public static int CalculateMinimumVersion(int length, EncodingMode encMode, ECCL
             }
 
             // if no version was found, throw an exception
-            var maxSizeByte = _capacityTable.Where(
-                x => x.Details.Any(
```
gfoidl, line 98: This closure also showed up, and unfortunately the closure (display-class) gets created at method entry, and not when actually needed, so we have to work around this.

codebude:
Hi @gfoidl ,  

gfoidl:
As written in https://github.com/codebude/QRCoder/issues/605#issuecomment-3342962313 I'd like that the repo here continues to live, and don't be archived, so please merge the PR as it brings goodness to your baby QRCoder.

Shane32:
@gfoidl if you merge in master, tests should run properly now

codebude (approved):

PR #611, opened 2025-10-01
Optimize compression/decompression and add tests

Shane32:
## Summary by CodeRabbit

- New Features
  - More reliable save/load of QR code data across Uncompressed, Deflate, and GZip modes with header and size validation.

- Refactor
  - Switched to streaming-based QR data serialization to reduce memory and improve performance.
  - Added early validation and stronger error handling; consistent version/module parsing for Micro and Standard QR.

- Tests
  - Added round-trip tests verifying saved QR data reloads with identical version and module matrix across compression modes.

- Chores
  - Updated C# language version for the build.

### DIFF
--- a/QRCoder/QRCodeData.cs
+++ b/QRCoder/QRCodeData.cs
@@ -142,7 +142,11 @@ public class QRCodeData : IDisposable
         try
         {
             //Add header - signature ("QRR")
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
+            targetStream.Write([0x51, 0x52, 0x52, 0x00]);
+#else
             targetStream.Write(new byte[] { 0x51, 0x52, 0x52, 0x00 }, 0, 4);
+#endif
 
             //Add header - rowsize
             targetStream.WriteByte((byte)ModuleMatrix.Count);
--- a/QRCoder/QRCodeData.cs
+++ b/QRCoder/QRCodeData.cs
@@ -55,46 +55,48 @@ public class QRCodeData : IDisposable
     /// <summary>
     /// Initializes a new instance of the <see cref="QRCodeData"/> class with raw data and compression mode.
     /// </summary>
-    /// <param name="rawData">The raw data of the QR code.</param>
+    /// <param name="bytes">The raw data of the QR code.</param>
     /// <param name="compressMode">The compression mode used for the raw data.</param>
-    public QRCodeData(byte[] rawData, Compression compressMode)
+    public QRCodeData(byte[] bytes, Compression compressMode)
     {
-        var bytes = new List<byte>(rawData);
-
         //Decompress
         if (compressMode == Compression.Deflate)
         {
-            using var input = new MemoryStream(bytes.ToArray());
+            using var input = new MemoryStream(bytes);
             using var output = new MemoryStream();
             using (var dstream = new DeflateStream(input, CompressionMode.Decompress))
             {
                 dstream.CopyTo(output);
             }
-            bytes = new List<byte>(output.ToArray());
+            bytes = output.ToArray();
         }
         else if (compressMode == Compression.GZip)
         {
-            using var input = new MemoryStream(bytes.ToArray());
+            using var input = new MemoryStream(bytes);
             using var output = new MemoryStream();
             using (var dstream = new GZipStream(input, CompressionMode.Decompress))
             {
                 dstream.CopyTo(output);
             }
-            bytes = new List<byte>(output.ToArray());
+            bytes = output.ToArray();
         }
 
+        var count = bytes.Length;
+
+        if (count < 5)
+            throw new Exception("Invalid raw data file. File too short.");
         if (bytes[0] != 0x51 || bytes[1] != 0x52 || bytes[2] != 0x52)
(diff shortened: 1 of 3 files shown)


QRCoder/QRCodeData.cs
```diff
@@ -59,42 +59,44 @@ public QRCodeData(string pathToRawData, Compression compressMode) : this(File.Re
...
                 dstream.CopyTo(output);
             }
-            bytes = new List<byte>(output.ToArray());
+            rawData = output.ToArray();
         }
 
-        if (bytes[0] != 0x51 || bytes[1] != 0x52 || bytes[2] != 0x52)
+        var count = rawData.Length;
+
+        if (count < 5)
```
gfoidl:
```suggestion
        if (rawData.Length < 5)
```
To give the JIT some hints, so the bound checks in the array access below can be removed (at least on newer JIT versions), and there's no harm for older ones.

QRCoder/QRCodeData.cs
```diff
@@ -55,46 +55,48 @@ public QRCodeData(string pathToRawData, Compression compressMode) : this(File.Re
...
+            throw new Exception("Invalid raw data file. File too short.");
         if (bytes[0] != 0x51 || bytes[1] != 0x52 || bytes[2] != 0x52)
             throw new Exception("Invalid raw data file. Filetype doesn't match \"QRR\".");
 
         //Set QR code version
         var sideLen = (int)bytes[4];
-        bytes.RemoveRange(0, 5);
         Version = (sideLen - 21 - 8) / 4 + 1;
```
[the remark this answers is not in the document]
gfoidl: My knowledge about QR is too little to judge, but at least it reads coherent...

QRCoder/QRCodeData.cs
```diff
@@ -121,60 +122,64 @@ public QRCodeData(byte[] rawData, Compression compressMode)
...
+            var dataQueue = new Queue<int>();
+            foreach (var row in ModuleMatrix)
             {
-                b += (byte)(dataQueue.Dequeue() << i);
+                foreach (var module in row)
+                {
+                    dataQueue.Enqueue((bool)module ? 1 : 0);
+                }
+            }
+            for (int i = 0; i < 8 - (ModuleMatrix.Count * ModuleMatrix.Count) % 8; i++)
```
gfoidl:
```suggestion
            int mod = (int)(((uint)ModuleMatrix.Count * (uint)ModuleMatrix.Count) % 8);
            for (int i = 0; i < 8 - mod; i++)
```
I'm not sure the JIT will keep the expression in a register. Re-evaluation is quite expensive, by multiplicaiton and especially the division (for the modulus).
So compute once as `uint`, as then the mod becomes a cheap bit-shift too.

QRCoder/QRCodeData.cs
```diff
@@ -121,60 +132,65 @@ public QRCodeData(byte[] rawData, Compression compressMode)
...
         }
 
-        //Process queue
-        while (dataQueue.Count > 0)
+        try
         {
-            byte b = 0;
-            for (int i = 7; i >= 0; i--)
+            //Add header - signature ("QRR")
+            targetStream.Write(new byte[] { 0x51, 0x52, 0x52, 0x00 }, 0, 4);
```
gfoidl, line 158: Let's use collection expression to avoid the `byte[]` allocation for .NET 8+ (where they were introduced).
```suggestion
#if NET8_OR_GREATER
            targetStream.Write([0x51, 0x52, 0x52, 0x00]);
#else
            targetStream.Write(new byte[] { 0x51, 0x52, 0x52, 0x00 }, 0, 4);
#endif
```

I thinks .NET 8+ is finde, as older targets aren't supported anymore by .NET.
Else, we could keep the signature in a static readonly and load it from there. But not so nice code...

Also .NET 10's JIT would stack allocate that array, so no heap allocation.
But the collection expression reads much nicer.

gfoidl (commented):
Forgot to send that comment before.

QRCoder/QRCodeData.cs
Shane32: I removed count and replaced with `rawData.Length` all 3 times - please verify.

QRCoder/QRCodeData.cs
Shane32: I bumped the C# version to 12, so the code you supplied will work as far back as .NET Core 2.1 -- anything supported by Span.  I'll update the conditional again to HAS_SPAN in a future PR, and probably add netstandard2_1 as a target at the same time.

QRCoder/QRCodeData.cs
```diff
@@ -152,7 +152,11 @@ public byte[] GetRawData(Compression compressMode)
         try
         {
             //Add header - signature ("QRR")
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
+            targetStream.Write([0x51, 0x52, 0x52, 0x00]);
```
gfoidl, line 156: Just as note: for .NET less than 8 this will still allocate the array, as the collection builder, etc. isn't available in .NET, so Roslyn can't use these types and produces "fallback-code". 
The change is still good, as it clearer to read. Thanks.

gfoidl (approved):

PR #613, opened 2025-10-01
Fix #if conditional to match rest of application

Shane32:
## Summary by CodeRabbit

* **Refactor**
  * Optimized numeric QR encoding to reduce allocations and improve speed on newer runtimes by using more efficient parsing for fixed-width digit groups.
  * Enhances performance when generating QR codes with numeric content, especially for large inputs, while preserving identical output and compatibility across supported platforms.

### DIFF
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -1009,7 +1009,7 @@ public partial class QRCodeGenerator : IDisposable
         for (int i = 0; i < plainText.Length - 2; i += 3)
         {
             // Parse the next three characters as a decimal integer.
-#if NET5_0_OR_GREATER
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
             var dec = int.Parse(plainText.AsSpan(i, 3), NumberStyles.None, CultureInfo.InvariantCulture);
 #else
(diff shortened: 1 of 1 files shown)


QRCoder/QRCodeGenerator.cs
```diff
@@ -1009,7 +1009,7 @@ private static BitArray PlainTextToBinaryNumeric(string plainText)
         for (int i = 0; i < plainText.Length - 2; i += 3)
         {
             // Parse the next three characters as a decimal integer.
-#if NET5_0_OR_GREATER
+#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1
```
gfoidl, line 1012:
```suggestion
#if NETCOREAPP || NETSTANDARD2_1
```
I thinks this could be shortened for better legibility.
Older .NET Core targets than 2.1 aren't present. And .NET (5+) is covered by that compiler constant too.

Shane32: Well, as written:
- it matches the rest of the codebase
- it matches the conditions for which Span is supported

So I'd prefer to leave it as-is here for those reasons.  Perhaps it would be better to have a compile time constant defined for span support - similar to SYSTEM_DRAWING we have SPAN - then in the csproj we define it for the applicable target frameworks.  Then we change all uses to match.

gfoidl (approved):
Your suggestion with a `#define HAS_SPAN` (or similar) sound reasonable as follow-up work.

PR #614, opened 2025-10-01
Update testing to only test .NET Standard and LTS versions

Shane32:
Changes:
- Remove .NET Framework 3.5 testing (which was running in 4.0 compatibility mode anyway)
- Remove .NET Framework 4.5.2 testing
- Add .NET Framework 4.6.2 testing (which is currently supported by MS)
- Remove .NET Core 1.1 testing
- Remove .NET Core 2.0 testing
- Add .NET Core 2.1 LTS testing
- Add .NET Core 3.1 LTS testing
- Refactor assets to be embedded in the test dll
- Simplify csproj dependencies and configuration

This means:
- The .NET Framework 3.5 build is not tested anymore
- The .NET Standard 1.3 build is not tested anymore
- The .NET 4.0 build should still be tested via 4.6.2 testing
- The .NET Standard 2.0 build should be tested via Core 2.1 and Core 3.1 testing

## Summary by CodeRabbit

- New Features
  - No user-facing changes in this release.

- Tests
  - Modernized test targets to newer .NET versions and removed legacy ones.
  - Switched tests to use embedded logo assets instead of file system access.
  - Simplified culture handling and conditional compilation in tests.
  - Updated test dependencies and consolidated package versions.

- Chores
  - Updated CI workflow matrix to align with supported frameworks.
  - Cleaned up solution items for maintainability.

### DIFF
--- a/.github/workflows/wf-build-test.yml
+++ b/.github/workflows/wf-build-test.yml
@@ -36,10 +36,9 @@ jobs:
     strategy:
       matrix:
         dotnet-framework:
-          - { name: ".NET 3.5", framework: "net35", coverage: false, no-build: true, sdk: "8.0.x" }
-          - { name: ".NET 4.52", framework: "net452", coverage: false, no-build: true, sdk: "8.0.x" }
-          - { name: ".NET Core 1.1", framework: "netcoreapp1.1", coverage: false, no-build: false, sdk: "1.0.x" }
-          - { name: ".NET Core 2.0", framework: "netcoreapp2.0", coverage: true, no-build: false, sdk: "2.0.x" }
+          - { name: ".NET Framework 4.6.2", framework: "net462", coverage: false, no-build: true, sdk: "8.0.x" }
+          - { name: ".NET Core 2.1", framework: "netcoreapp2.1", coverage: true, no-build: true, sdk: "2.1.x" }
+          - { name: ".NET Core 3.1", framework: "netcoreapp3.1", coverage: true, no-build: true, sdk: "3.1.x" }
           - { name: ".NET 5.0", framework: "net5.0", coverage: true, no-build: true, sdk: "5.0.x" }
           - { name: ".NET 5.0 Windows", framework: "net5.0-windows", coverage: true, no-build: true, sdk: "5.0.x" }
           - { name: ".NET 6.0", framework: "net6.0", coverage: true, no-build: true, sdk: "6.0.x" }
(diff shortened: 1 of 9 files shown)


Shane32:
Once merged, #611 can be modified to use `[InlineData]` for its added tests across all tested frameworks

Shane32:
@gfoidl I'd like your input here.  This PR should simplify testing libraries - now we can use any .NET Standard 2.0 libraries within the test project - but does mean that we can't test the .NET Framework 3.5 and .NET Standard 1.3 builds. Thoughts?

My thought is that these are so far outdated anyway it doesn't matter.  If someone runs into a bug on one of these frameworks, they can report it.  And I'd drop these frameworks for the next major version.  Users can continue to use QR Coder 1.x if they need .NET Framework 3.5 or .NET Core 1.x support

gfoidl (approved):
I'm always in favor of simplification. 
.NET Standard 2.0 as baseline is sufficient. So I'm fine dropping the older test-targets.

> drop these frameworks for the next major version.
👍🏻 

For the current changes incl. changes in tests, we should increase at least the minor-version, to have some clear(er) distinction.

Shane32:
> For the current changes incl. changes in tests, we should increase at least the minor-version, to have some clear(er) distinction.

Agree.  Typically I follow this pattern for other open source projects:
- Major version for breaking changes
- Minor version for new features / new API (applies now due to micro QR code support)
- Revision for bug fixes

Generally I also allow bug fixes that could be considered a breaking change from an atypical viewpoint, but depending on the scenario, I may trigger a minor version bump in that scenario.  Hard to give an example; it's pretty rare.  In that scenario I'd point it out in the PR for you to review.

Any code path that will trigger an exception (especially where it would always trigger an exception) is typically not a breaking change either.  For instance, these scenarios are never considered breaking:
- Changes to exception messages
- Adding exceptions to protect against code that would fail anyway (e.g. adding an `ArgumentNullException` to protect against `NullReferenceException`)
- Using a derived class for an exception where the inherited class was previously triggered (e.g. using `BadQrCodeException : InvalidOperationException` instead of `InvalidOperationException`)

I try never to add dependencies within a major version, but we can discuss if there's a reasonable reason to do so.  I try to only bump dependencies across minor versions where it is clear that the updated dependency does not place additional constraints on the developer -- and never bump dependencies across revisions.  For example, if we target .NET 8, perhaps we bump System.Drawing.Common to the .NET 8 version - but only for the .NET 8 target, keeping the dependency as it was for other targets.

We can discuss for anything unclear.

gfoidl:
> We can discuss for anything unclear.

Nothing unclear. This matches quite well how I'd like to do versioning.

Maybe extract your last comment into a `versioning-policy.md`, so that others know how it's handled here.

PR #622, opened 2025-10-02
Fix: Make BIC optional for Girocode Version 2

### LINKED ISSUE #582: Girocode / PayloadGenerator, BIC validation errors when not passed in to Girocode constructor, with SCT version set to 2
## Type of issue
[x] Bug  

## Expected Behavior
A new instance of **Girocode** should not throw a nullrefexception or return validation errors when **BIC** is not supplied and **SCT version** is set to _Girocode.GirocodeVersion.Version2_.

## Current Behavior
BIC is validated regardless of _Girocode.GirocodeVersion_. When not supplied a null ref occurs on 
`_iban = iban.Replace(" ", "").ToUpper();`

Always validate BIC when _Girocode.GirocodeVersion.Version1_.
Only validate BIC when _Girocode.GirocodeVersion.Version2_ and BIC is supplied.

`var payload = new Girocode(iban:"NL86INGB0002445588", bic: null, name: "a name", remittanceInformation: "some info", amount: 1337.99m, version: Girocode.GirocodeVersion.Version2, encoding: Girocode.GirocodeEncoding.UTF_8);`

## Your Environment

* Version used:  1.6.0
* Compiled from source or NuGet package?:  NuGet package
* Payload/Encoded text:  see example
* Used payload generator:  Girocode
* Used ECC-level:  15
* Used renderer class:  Base64QRCode
* Environment (.NET 3.5, .NET 4.X, .NETCore, ...):  NET8

Shane32:
Roo: I've successfully fixed the GitHub issue and completed all requested tasks:

### Fixed Issue
The Girocode constructor was throwing a `NullReferenceException` when BIC was null with Version 2, which should be allowed per the Girocode specification.

### Changes Made

1. **Modified [`IsValidBic()`](QRCoder/PayloadGenerator.cs:70)** in PayloadGenerator.cs:
   - Added an overload that accepts a `required` parameter
   - Returns `true` for null/empty BIC when not required
   - Returns `false` for null/empty BIC when required

2. **Updated [`Girocode` constructor](QRCoder/PayloadGenerator/Girocode.cs:48)** in Girocode.cs:
   - BIC validation now checks if it's required based on version (Version 1 requires BIC, Version 2 makes it optional)
   - Uses null-conditional operator (`?.`) and null-coalescing operator (`??`) for clean BIC handling

3. **Added comprehensive tests** in GirocodeTests.cs:
   - Version 2 with null BIC (passes)
   - Version 2 with empty BIC (passes)
   - Version 2 with valid BIC (passes)
   - Version 2 with invalid BIC (throws exception)
   - Version 1 with null BIC (throws exception)
   - Version 1 with empty BIC (throws exception)

### Test Results
All 23 Girocode tests passed successfully across all target frameworks (net462, netcoreapp2.1, netcoreapp3.1, net5.0, net5.0-windows, net6.0, net6.0-windows).

### Branch and Commit
- Created branch: `fix/girocode-version2-bic-optional`
- Committed changes with detailed commit message
- Pushed to remote repository

You can create a pull request at: https://github.com/Shane32/QRCoder/pull/new/fix/girocode-version2-bic-optional

## Summary by CodeRabbit

* New Features
  * Girocode v2 now accepts omitted BIC values while Girocode v1 continues to require a BIC.
* Bug Fixes
  * Safer handling of null/empty BIC inputs and clearer validation behavior that varies by Girocode version.
* Tests
  * Added unit tests covering presence, absence, and invalid BIC cases and related payload/output behavior.


QRCoder/PayloadGenerator/Girocode.cs
```diff
@@ -36,16 +36,16 @@ public class Girocode : Payload
...
             _version = version;
             _encoding = encoding;
             if (!IsValidIban(iban))
                 throw new GirocodeException("The IBAN entered isn't valid.");
             _iban = iban.Replace(" ", "").ToUpper();
-            if (!IsValidBic(bic))
+            if (!IsValidBic(bic, _version == GirocodeVersion.Version1))
                 throw new GirocodeException("The BIC entered isn't valid.");
-            _bic = bic.Replace(" ", "").ToUpper();
+            _bic = bic?.Replace(" ", "").ToUpper() ?? string.Empty;
```
gfoidl, line 48:
```suggestion
            _bic = bic?.Replace(" ", "").ToUpper() ?? "";
```
To a huge part a matter of taste, but using the literal `""` has a few advantages over using `string.Empty` (in general, not specifi to this repo):
* less IL code as the literal is a constant as opposed a field load -> smaller assembly
* less work for the JIT, as no load and inlining of that field is needed

If a change is done, we should it do for the whole project to be consistent.
If not, I'm also fine with it, as these two listed bullet don't change much here.

This is more a general note.

QRCoder/PayloadGenerator.cs
```diff
@@ -65,6 +65,18 @@ private static bool IsValidQRIban(string iban)
...
+    /// Validates the structure of a BIC with optional requirement check.
+    /// </summary>
+    /// <param name="bic">The BIC to validate.</param>
+    /// <param name="required">Whether the BIC is required. If false, null/empty values are considered valid.</param>
+    /// <returns>True if the BIC is valid; otherwise, false.</returns>
+    private static bool IsValidBic(string? bic, bool required)
+    {
+        if (string.IsNullOrEmpty(bic))
+            return !required;
+        return IsValidBic(bic!);
```
gfoidl, line 78:
```suggestion
        return IsValidBic(bic);
```
Null analysis in Roslyn should know that at this point `bic` isn't null.

gfoidl (approved):
Modulo the nits looks good to me.

QRCoder/PayloadGenerator.cs
Shane32: Unfortunately when targeting .net 3.5 and 4.0 the attributes are missing from IsNullOrEmpty so the compiler doesn't have that guarantee and the ! is necessary. (We have our own IsNullOrWhiteSpace function defined as a string extension to overcome this issue, but not IsNullOrEmpty.)

gfoidl: Ah, I forgot that the `[NotNullWhen(false)]` attribute isn't there on older targets and that the compiler uses -- of course -- that ref-assemblies. Thanks for the reminder!

QRCoder/PayloadGenerator/Girocode.cs
Shane32: In general, I'd like to follow .NET analysis rules. I'm not sure exactly what that means in this specific case, but one of my goals is to enable the analyzers at the recommended level. It might not be practical until we remove support for .NET Framework in v2...

Shane32: #638

Issue #240, opened 2020-05-15
[New Feature]


serfend:
## Type of issue
[ ] Bug  
[ ] Question (e.g. about handling/usage)  
[*] Request for new feature/improvement  

## Expected Behavior
when use  `new QRCode(qrCodeData).GetGraphic()` to generate a code , it would be too slow to joying users when a large pixelsPerModule is set.
it'd be better to use qrcode render to get o(1) rather than use for(for())g.FillRectangle() on o(n²).

Issue #489, opened 2024-03-28
QRCode namespace doesn't exist in .NET 8


ScottDurkin:
## Type of issue
[X ] Bug  
[ ] Question (e.g. about handling/usage)  
[ ] Request for new feature/improvement  

## Expected Behavior
Should be able to create new QRCode Object by passing in QRCodeData.

## Current Behavior
QRCode namespace doesn't exist.

## Steps to Reproduce (for bugs)
Upgrade to .NET Core 8

## Your Environment
Visual Studio 2022 latest version with .NET Core 8. 

Code sample that works in .NET Core 5
` QRCodeGenerator qrGenerator = new QRCodeGenerator();
 QRCodeData qrCodeData = qrGenerator.CreateQrCode(PublicProfileURL, QRCodeGenerator.ECCLevel.Q);
 QRCode qrCode = new QRCode(qrCodeData);
 Bitmap qrCodeImage = qrCode.GetGraphic(20);`

ScottDurkin:
Note, the above code does not work in either .NET Core 6 OR 7.

cimpok:
Hi @ScottDurkin ,
I am having the same issue since a long time, for me as a workaround it worked downgrading the package to 1.4.1

cezar-pimentel:
Yeah, for me too. I believe the problem was created with the 1.4.3 changes.
They'll need to add net7 and net8 as target frameworks as well.
@codebude, would you like me to create a PR to address this issue?

ScottDurkin:
> Hi @ScottDurkin ,
> 
> I am having the same issue since a long time, for me as a workaround it worked downgrading the package to 1.4.1

Thank you for your response! I'll take a look!

codebude:
@cezar-pimentel Thanks for your offer. I would love to see such PR. But currently I'm going through all the older PRs and see if I can merge them. I fear if you start your PR now, the master will change a couple of times, while you're working on the PR. Maybe it's better to wait until the majority of the older PRs is merged.

cezar-pimentel:
> @cezar-pimentel Thanks for your offer. I would love to see such PR. But currently I'm going through all the older PRs and see if I can merge them. I fear if you start your PR now, the master will change a couple of times, while you're working on the PR. Maybe it's better to wait until the majority of the older PRs is merged.

No worries, we can wait. There's no need to rush.

Thank you my friend! 👊💪

alban-ameti:
[...]

PR #626, opened 2025-10-02
optimise: fill `QRCode` with one colour

TimothyMakkison:
Fill bitmap with the light colour and then draw with the black, halving the number of updates.

Would love to benchmark this, but I can't get the benchmark file to use `System.Drawing` to reference `QRCoder.QRCode`. I've tried setting the .NET version of the benchmarks to `6.0.0` and referencing `System.Drawing.Common` but I can't reference `QRCoder.QRCode`

Edit: I can do this in other places

Related #240

## Summary by CodeRabbit

* **Refactor**
  * Improved QR code rendering by pre-filling the background and drawing only dark modules, reducing per-pixel work.
  * Faster and more responsive rendering, especially for large or high‑DPI outputs.
  * Visual output for valid QR codes remains unchanged and no public interfaces were modified.


Shane32:
Try changing the benchmark project's target framework to `net8.0-windows`

Ref:
- #489

Shane32:
FYI, it may be possible to create a single path for the dark modules, issuing a single Fill call, further improving performance.  Just guessing, and not sure what the performance impact would be.

https://chatgpt.com/s/t_68ded921c014819186a24654c04b167a

TimothyMakkison:
> net8.0-windows

Thanks 🙏 

> FYI, it may be possible to create a single path for the dark modules, issuing a single Fill call, further improving performance. Just guessing, and not sure what the performance impact would be.

> https://chatgpt.com/s/t_68ded921c014819186a24654c04b167a

Yeah, I wanted to use benchmarks to test this. I'd also like to know if it's worth filling the bitmap with the light or dark colour or how much faster `gfx.Clear` is vs painting a large rectangle. 

IIRC there was an issue with mutating a bitmap a lot in C#, I can't remember if it was related to `bmp.LockBits` and unlocking. I did wonder if it would be better to write to a scale bitmap and then scale it up to the correct size `Bitmap resized = new Bitmap(original,new Size(original.Width * 4,original.Height*4));`

I'll experiment more in the morning

### After

| Method                       | Mean         | Error       | StdDev       | Allocated |
|----------------------------- |-------------:|------------:|-------------:|----------:|
| DrawingBitmapByteQRCodeSmall |     855.7 us |    16.95 us |     32.25 us |     313 B |
| DrawingBitmapByteQRCodeMedium|   2,781.1 us |    55.02 us |    127.52 us |     316 B |
| DrawingBitmapByteQRCodeBig   |  46,853.3 us |   968.01 us |  2,649.90 us |     401 B |
| DrawingBitmapByteQRCodeHuge  | 325,011.8 us | 6,481.17 us | 12,793.20 us |     848 B |


### Before
| Method                        | Mean       | Error      | StdDev     | Allocated |
|------------------------------ |-----------:|-----------:|-----------:|----------:|
| DrawingBitmapByteQRCodeSmall  |   2.411 ms |  0.0481 ms |  0.0843 ms |     316 B |
| DrawingBitmapByteQRCodeMedium |   6.795 ms |  0.1246 ms |  0.2545 ms |     320 B |
| DrawingBitmapByteQRCodeBig    |  90.470 ms |  1.7894 ms |  4.0023 ms |     491 B |
| DrawingBitmapByteQRCodeHuge   | 521.257 ms | 10.3711 ms | 18.9641 ms |    1384 B |

I can see why #240 was created, half a second for a large QR code is crazy 0_0, not sure why the memory usage is so low for a huge bitmap

Shane32 (approved):

PR #635, opened 2025-10-05
Change PdfByteQRCode to draw rectangles instead of embed JPGs

Shane32:
This changes the PdfByteQRCode algorithm to draw rectangles rather than to embed an image.  This results in a PDF that has no compression artifacts and is about 50% the size of the previous JPG-embedded image (at default compression settings).  It also allows the renderer to be compatible with any OS.

As an added benefit, the previous rendering process would convert to PNG first (using PngByteQRCode), then read it via GDI+, then save it via GDI+ to JPEG.  All of that has been eliminated, making this code much faster.  I did not perform any string optimizations within this PR as it was not within scope.

Extensive comments to the PDF commands were added so it is understood what the code actually does.

The jpgQuality parameter is now unused.  While I could add a method without the parameter and mark the old method as obsolete, this would technically be a source-breaking change (not binary-breaking) in certain cases.  So for v1.x I did not change it.

## Summary by CodeRabbit

* **New Features**
  * Generate QR codes directly as PDF bytes with vector-based rendering for crisp, scalable output.
  * Customizable dark/light colors and DPI; parameterless instantiation supported.
  * Cross-platform support (Windows-only restriction removed).

* **Changes**
  * Simplified API: removed obsolete jpgQuality parameter; standardized GetGraphic signature and streamlined PDF output.

* **Tests**
  * Added tests for black/white, color, custom DPI, parameterless usage, and helper-based PDF generation.


Shane32:
For better readability of the GitHub diff, please enable 'hide whitespace'.
image

QRCoder/PdfByteQRCode.cs
```diff
@@ -58,167 +51,205 @@ private byte[] HexColorToByteArray(string colorString)
...
+    /// <param name="color">RGB color as byte array.</param>
+    /// <returns>PDF color string (three decimal values).</returns>
+    private static string ColorToPdfRgb(byte[] color)
+    {
+        if (color.Length != 3)
+            throw new ArgumentException("Color must be a 3-byte RGB array", nameof(color));
+
+        var r = ToStr(color[0] / (float)255);
+        var g = ToStr(color[1] / (float)255);
+        var b = ToStr(color[2] / (float)255);
```
gfoidl: Just nice to have, for a tad better perf:
```suggestion
        const float inv255 = 1 / 255f;
        var r = ToStr(color[0] * inv255);
        var g = ToStr(color[1] * inv255);
        var b = ToStr(color[2] * inv255);
```

In strict term for floating point operations, that transform is not 100 % correct, but the error should be towards 0, so no real problem.

gfoidl (approved):
Really nice 👍🏻 
And thanks for all the comments, to make it easier to understand and for better future maintenance.

PR #639, opened 2025-10-05
Reference System.Drawing.Common for all targets

Shane32:
This PR is currently opened only for review/discussion.

## Summary by CodeRabbit

- New Features
  - CI now performs a full solution build in key workflows to catch build issues earlier.

- Chores
  - Simplified .NET target configuration to reduce platform-specific targets and unify behavior.
  - Consolidated drawing-library dependency handling for more consistent behavior across supported frameworks.
  - Adjusted test project targeting to align with Windows-focused test runs.


Shane32:
@gfoidl I'm reconsidering this change (meaning, I'm considering merging this PR).  If we look at #489 we see a great number of issues raised due to people not understanding why the various classes are "missing" in .NET 6+.  And these are just the people that couldn't find the solution and created an issue; many more people likely had the same problem but solved it themselves.  Further, I've needed to help yet another person in the first week I've been maintaining the project.

It just seems to me to be simpler for everyone to include the reference to System.Drawing.Common for all target frameworks, letting the .NET Analyzers and/or MS exceptions inform developers if they try to use it on Linux.

If someone really wants to omit the System.Drawing.Common reference on Linux, they can stick with QRCoder 1.6.0, or use v2.  But if we merge this, it will be binary-compatible with v1.4.x, the lack of which is a common cause of complaint.

Comments?

gfoidl (approved):
OK, when it becomes less churn for the community it's good.

PR #640, opened 2025-10-06
Enable recommended analyzers; use invariant culture throughout

Shane32:
## Summary by CodeRabbit

- Improvements
  - Consistent culture-invariant formatting for amounts, dates, and casing across payloads (IBAN/BIC, Girocode, BezahlCode, Swiss QR, Monero, Calendar, Contact).
- Deprecations
  - BezahlCode legacy authority types now display explicit guidance directing to SEPA-compliant alternatives.


QRCoder/Extensions/StringExtensions.cs
```diff
@@ -29,4 +29,30 @@ public static bool IsNullOrWhiteSpace(
         return string.IsNullOrWhiteSpace(value);
 #endif
     }
+
+    /// <summary>
+    /// Converts a hex color string to a byte array.
+    /// </summary>
+    /// <param name="colorString">Color in HEX format like #ffffff.</param>
+    /// <returns>Returns the color as a byte array.</returns>
+    internal static byte[] HexColorToByteArray(this string colorString)
```
gfoidl, line 38: Instead of the `byte[]` could we use a 
```c#
public struct Color
{
    public byte R;
    public byte G;
    public byte B;
    public byte A;
}
```
to avoid the (temp.) allocation?

Or when the array-like semantics should be kept, an InlineArray (where supported) / a fixed sized buffer when not supported?

Could be done in a separate PR also. Maybe that would be event better, as such "helper types" can be created there, and with a pass over the code base more such occurances be changed.

QRCoder/PayloadGenerator/Girocode.cs
```diff
@@ -81,7 +81,7 @@ public override string ToString()
             girocodePayload += _bic + _br;
             girocodePayload += _name + _br;
             girocodePayload += _iban + _br;
-            girocodePayload += $"EUR{_amount:0.00}".Replace(",", ".") + _br;
+            girocodePayload += string.Format(CultureInfo.InvariantCulture, "EUR{0:0.00}", _amount) + _br;
```
gfoidl, line 84: There are lots of intermediate allocations, that could be avoided.

For .NET 10 the JIT will probably elide them, but for .NET 6+ string interpolation should be used. And IIRC for older targets only some attributes need to be polyfilled, so that the compiler can emit the proper code.

gfoidl (approved):
We can look at the suggestions later.
Anyway I'll do a optimization pass accross the code base when I have a bit more time.

QRCoder/Extensions/StringExtensions.cs
Shane32: #647

QRCoder/PayloadGenerator/Girocode.cs
Shane32: #648

PR #641, opened 2025-10-06
Fix CI tests for most frameworks

Shane32:


Shane32:
@gfoidl I'm still fighting CI scripts at every turn.  Apparently only the .NET Core 2.1 tests were running previously.  I think I got it this time.  Lmk if you have any comments.  I'm merging so I can run tests on my other PRs.

gfoidl:
Unfortunately for CI the most reliable way is to just try it live. So far I don't know a good way to test it in DEV environment.

PR #643, opened 2025-10-06
Make Postscript renderer deterministic

Shane32:
Removes the optional creation date from the created postscript instructions.  Should have no effect on output.  Makes it easier for developers that want to write tests for their code that uses QRCoder to create postscript output.  Also converted constants to raw string literals.

## Summary by CodeRabbit

- Bug Fixes
  - Removed timestamp from PostScript/EPS headers for stable, reproducible output across runs and platforms.

- Refactor
  - Streamlined PostScript generation internals without altering behavior or public APIs.

- Tests
  - Updated approval tests to align with deterministic PostScript/EPS output.

### DIFF
--- a/QRCoder/PostscriptQRCode.cs
+++ b/QRCoder/PostscriptQRCode.cs
@@ -101,7 +101,7 @@ public class PostscriptQRCode : AbstractQRCode, IDisposable
         var pointsPerModule = (double)Math.Min(viewBox.Width, viewBox.Height) / (double)drawableModulesCount;
 
         string psFile = string.Format(PS_HEADER, new object[] {
-            DateTime.Now.ToString("s"), CleanSvgVal(viewBox.Width), CleanSvgVal(pointsPerModule),
+            CleanSvgVal(viewBox.Width), CleanSvgVal(pointsPerModule),
             epsFormat ? "EPSF-3.0" : string.Empty
         });
         psFile += string.Format(PS_FUNCTIONS, new object[] {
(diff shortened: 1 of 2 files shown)


QRCoder/PostscriptQRCode.cs
```diff
@@ -130,68 +130,75 @@ public string GetGraphic(Size viewBox, Color darkColor, Color lightColor, bool d
     /// <returns>Returns the cleaned string representation of the double value.</returns>
     private string CleanSvgVal(double input) => input.ToString(System.Globalization.CultureInfo.InvariantCulture);
 
-    private const string PS_HEADER = @"%!PS-Adobe-3.0 {3}
-%%Creator: QRCoder.NET
-%%Title: QRCode
-%%CreationDate: {0}
```
Shane32, line 136: This "CreationDate" line is the only removed line; the constant is otherwise identical.  (although {1} becomes {0} etc)

QRCoderTests/PostscriptQRCodeRendererTests.cs
```diff
@@ -9,7 +9,7 @@ public void can_render_postscript_qrcode_simple()
         var gen = new QRCodeGenerator();
         var data = gen.CreateQrCode("This is a quick test! 123#?", QRCodeGenerator.ECCLevel.L);
         var ps = new PostscriptQRCode(data).GetGraphic(5);
-        RemoveCreationDate(ps).ShouldMatchApproved(x => x.NoDiff());
+        ps.ShouldMatchApproved(x => x.NoDiff());
```
Shane32, line 12: Tests previously removed the creation date from the verified files.  So all tests continue to match with no change to test result data.

gfoidl (approved):

PR #644, opened 2025-10-06
Add transposition tests

Shane32:
The ASCII "small" renderer is currently rendering QR codes transposed along the x/y diagonal.  This suite of tests will help to catch these problems.

As many as possible of the renderers (png, base64, bmp, Bitmap, svg) will compare against the same reference image; in so doing, it will be easier to verify that changes do not cause transposition, and that someone does not accidentally save a transposed image as valid.  However, some renderers (art, pdf, postscript, ascii) will require human verification at this time.

All tests will render the exact same QR code, so any test can be simply verified against any other of the tests.

There is also a 'reference' image showing the correct black-module location; the transposed location is white.  Additionally, there is a test to ensure that the transposed location is white (since this is not enforced by the QR specification, but merely the chosen QR code is white there to discern from the black-module position).

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive verification tests to ensure QR codes are not transposed across multiple renderers and formats (ASCII, bitmap/base64/PNG, SVG, PDF, PostScript, Art).
  * Added approved ASCII snapshot resources in multiple sizes and image-based approval checks for visual consistency.
  * Added targeted checks for specific module positioning to prevent orientation regressions.

* **Chores**
  * Added an SVG rendering dependency to support cross-format validation in the test suite.


Shane32:
Waiting on #645 and #643 before this PR can be merged

gfoidl (approved):

PR #655, opened 2025-10-08
Add path-based SVG rendering with RLE encoding to reduce file size and memory usage

Shane32:
Changes:

- Uses paths to create a SVG versus many individual rectangles
- Supports transparent or partially-transparent colors - for both light and dark modules
- Added `GetGraphic()` overload with no arguments for a scalable SVG (no width/height set) with black on white
- Draws the light color as a background (assuming the dark color has no transparency)
- Omits drawing the light or dark color if they are completely transparent
- Draws the light color behind logos, as was done with the previous algorithm
- 70% file size reduction or so
- Numbers are all encoded with invariant culture
- Considerable memory reduction
  - Only main StringBuilder allocates memory (assuming no logo in use), and final `stringBuilder.ToString()`
  - No matrix for 2d compression of rectangles
  - Encoding of numbers is done with a `stackalloc` and copied directly into the StringBuilder
- RLE encoding of rectangles per row
- Elimination of SVG v1.1 marker, an optional marker; generated SVG should be compatible with both 1.1 and 2.0 (unless the user specifies an RGBA color value besides the keyword `transparent`)

Todo:

- [x] Verify that the light color should be drawn behind logos
- [x] Add transparency tests

Notes:

1. `viewBox` argument no longer affects the actual SVG view box.  According to online documentation, the view box has no bearing whatsoever on the rendered size of the SVG.  The view box just tells the viewer (e.g. browser) what portion of an infinitely large 2-dimensional coordinate space within the SVG to display.  Only the `width` and `height` SVG tag parameters affect the displayed size.  The only way this would affect a user/developer is if they were altering the SVG content after its generation.  It should not affect embedding the SVG inside another SVG, as the embedded SVG coordinate system is its own.

2. Also, the RLE encoding in a single dimension replaced a 2-dimensional rectangle detection algorithm previously in use.  I find the previous algorithm to have very little usefulness beyond the finder boxes.  In the future, it would be best to have an algorithm that would encode adjoining modules of any shape, which is certainly possible using even-odd fill.  (Applies to PDF writing as well.)  But it would come at the cost of additional memory requirements.

3. The rendered SVG within the transposition tests still passes, since the SVG when rendered to a bitmap has not changed with this PR.  Visual inspection of all prior tests indicate that the SVG produces identical output visually.

## Summary by CodeRabbit

* **New Features**
  * Added a parameterless SVG export, explicit sizing modes, hex color inputs, and improved logo placement and obstruction handling.

* **Refactor**
  * Streamlined SVG assembly and numeric formatting for more consistent, size-aware SVG output and transparency handling.

* **Chores**
  * Updated public documentation to reflect sizing, color, and logo behavior changes.

* **Tests**
  * Added tests for unscaled output, color/hex/ARGB inputs, transparency, and logo scenarios.


Shane32:
@gfoidl Some PRs are small ... some PRs are big ....  (lol)

QRCoder/SvgQRCode.cs
```diff
@@ -108,96 +117,85 @@ public string GetGraphic(Size viewBox, string darkColorHex, string lightColorHex
...
-        int[,] matrix = new int[drawableModulesCount, drawableModulesCount];
-        for (int yi = 0; yi < drawableModulesCount; yi += 1)
+        // Build SVG opening tag with size attributes
+        var svgFile = new StringBuilder();
+
+        svgFile.Append(@"<svg xmlns=""http://www.w3.org/2000/svg"" viewBox=""0 0 ");
+        Append(drawableModulesCount);
+        svgFile.Append(' ');
+        Append(drawableModulesCount);
+        svgFile.Append(@""" shape-rendering=""crispEdges""");
```
gfoidl: IMO this reads better. There are too many `"` for me 😉.
```suggestion
        svgFile.Append(@"\" shape-rendering=\"crispEdges\"");
```
(or using raw string literals).

I know that there are other places where it's fine as written, but here it starts and ends with 3 `"`, so my brain switched a bit into raw string literal mode, which is not correct.
W/ the suggestion it's clearer and easier to follow in this place.

QRCoder/SvgQRCode.cs
```diff
@@ -214,54 +212,229 @@ public string GetGraphic(Size viewBox, string darkColorHex, string lightColorHex
...
-            Y = imgPosY
-        };
+#if HAS_SPAN
+            Span<char> buffer = stackalloc char[16];
+            if (num.TryFormat(buffer, out int charsWritten, default, CultureInfo.InvariantCulture))
+            {
+                svgFile.Append(buffer.Slice(0, charsWritten));
+                return;
+            }
+#endif
```
gfoidl:
```suggestion
#if NET6_0_OR_GREATER
            svgFile.Append(CultureInfo.InvariantCulture, $"{num}");
#endif
```
Should also (w/o measuring) be a little bit faster, as it's formatted to the SB's buffer directly, and not via the span detour.

For < .NET 6 where span is available I wouldn't bother and don't have a separate #ifdef for this.

QRCoderTests/SvgQRCodeRendererTests.can_render_svg_qrcode.approved.svg
gfoidl, line 1: Well this savings are impressive 👍🏻

gfoidl (approved):
Just two minor things, otherwise LGTM.

Very impressive 😄.

QRCoder/SvgQRCode.cs
Shane32: I pushed this code into StringExtensions, with a fast path for .NET 6+, a fallback for HAS_SPAN, and fallback for anything else.  Your argument is completely valid; we really don't need to optimize for older targets....but why not - so long as it doesn't impede reading the code?  I think what I'd like to do is for v2 create an `Append(IFormatProvider, FormattableString)` method for pre-.NET6 environments and combine a bunch of these multi-line Append commands into a interpolated string for better readability and more concise code.

Shane32: I'd also like to consider a dependency on System.Memory for .NET Standard 2.0 support in v2 so that Span support is available throughout the codebase.  We can review a PR and see if it makes sense or not.  It may not help much if the framework (like int.Parse) does not support Span.

QRCoder/SvgQRCode.cs
Shane32: Changed all to use regular strings.  Raw strings are not possible when a string ends with a quote (unless using the multi-line variation).

QRCoder/PdfByteQRCode.cs
```diff
@@ -204,11 +204,11 @@ private string CreatePathFromModules()
...
                 // Format: x y width height re
-                pathCommands.Append(ToStr(startX));
+                pathCommands.AppendInvariant(startX);
                 pathCommands.Append(' ');
-                pathCommands.Append(ToStr(y));
+                pathCommands.AppendInvariant(y);
                 pathCommands.Append(' ');
-                pathCommands.Append(ToStr(x - startX));
+                pathCommands.AppendInvariant(x - startX);
                 pathCommands.Append(" 1 re\r\n");
```
Shane32, line 212: Hopefully for v2 this can all be cleaned up into an interpolated string, fully optimized for .NET Standard 2.1+, but compatible with .NET Standard 2.0 as well.

QRCoder/SvgQRCode.cs
gfoidl: > if the framework (like int.Parse) does not support Span.

I think that could be polyfilled with C#'s extension members. So a new compiler creaets code that runs on old targets.
Thus I also think that System.Memory makes the code base cleaner.

PR #672, opened 2025-10-13
Eliminates exceptions when characters are outside ISO-8859-1 range

Shane32:
@gfoidl  Although it only takes a few nanoseconds (billionths of a second) to run, it is 30% slower than the original code assuming all characters can be encoded.  On the other hand, it is a thousand times faster with zero allocations for a string that requires UTF-8.

| Method                   | Mean         | Error       | StdDev      | Gen0   | Allocated |
|------------------------- |-------------:|------------:|------------:|-------:|----------:|
| OriginalAsciiValid       |     6.222 ns |   0.4550 ns |   1.3415 ns |      - |         - |
| OriginalInvalidLastChar  | 9,625.746 ns | 190.4809 ns | 247.6790 ns | 0.1373 |     888 B |
| OptimizedAsciiValid      |     8.275 ns |   0.1907 ns |   0.2119 ns |      - |         - |
| OptimizedInvalidLastChar |    11.375 ns |   0.2461 ns |   0.2302 ns |      - |         - |

I could not find a more optimized approach.

### DIFF
--- a/QRCoder/QRCodeGenerator.cs
+++ b/QRCoder/QRCodeGenerator.cs
@@ -995,22 +995,21 @@ public partial class QRCodeGenerator : IDisposable
         bool IsUtf8() => (encoding == EncodingMode.Byte && (forceUtf8 || !IsValidISO(plainText)));
     }
 
-    private static readonly Encoding _iso88591ExceptionFallback = Encoding.GetEncoding(28591, new EncoderExceptionFallback(), new DecoderExceptionFallback()); // ISO-8859-1
     /// <summary>
     /// Checks if the given string can be accurately represented and retrieved in ISO-8859-1 encoding.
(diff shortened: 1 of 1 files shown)


gfoidl (approved):
Avoiding the potential allocation is better than trading the few nano seconds here.

Speeding this up could be done by vectorized code, but it gets ugly and long -- especially it can't be implemented for all TFMs.

.NET (in recent versions) provides `Ascii.IsValid` (and `Utf8.IsValid`), which use a vectorized approach, so we could check for valid ascii (the majority of inputs will be ascii) first, and when not fallback to the manual loop. In the worst case that's $O(2n)$.
But for what? Just a few ns in trade for complicated code...

I'm fine with the solution you have here.

PR #676, opened 2025-10-13
Add multi mode encoding

### LINKED ISSUE #585: Mixed Mode QR code
## Type of issue
[ ] Bug  
[X] Question (e.g. about handling/usage)  

## Expected Behavior

Hi, I just have question for how to generate a QR code.
I have requirement that the generated QR code should be Version 2 and ECC Q.
I know the limited data length is 29, but my data length is 30.
However, my data looks like this:  1X1234567890123456789012345:XX
in this case, I think i can split it to 3 different parts using different encode mode.
like: 1X -- Alpha Numeric
        1234567890123456789012345  -- Numeric
       :XX -- AlphaNumeric
 this should work in this case.

But I'm not sure is it possible to do this?

Shane32:
This PR includes a basic algorithm for producing multi-mode QR codes.  It is only activated when a QR code is created with default settings (apart from ECC level) and does not contain characters outside the Latin-1 character set.

The _benchmarks_ below for CreateQRCodeMultiMode demonstrate if the 'after' code produced the same version (size) QR code.  So in that example there is no net benefit to the user.

The _sample_ below shows how fewer bits encoded may produce a smaller QR code, which would both encode and render faster.

This PR does not include any code to automatically increase the ECC level if there are extra bits available to do so.

## Before

| Method                | Mean        | Error     | StdDev    | Gen0   | Allocated |
|---------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode          |    59.53 us |  1.181 us |  2.358 us | 0.1831 |   4.02 KB |
| CreateQRCodeMultiMode |   513.11 us |  2.108 us |  1.972 us |      - |   7.24 KB |
| CreateQRCodeLong      | 1,238.27 us |  9.495 us |  8.882 us |      - |   10.9 KB |
| CreateQRCodeLongest   | 6,810.61 us | 12.376 us | 11.577 us |      - |  43.84 KB |

## After

| Method                | Mean        | Error     | StdDev    | Gen0   | Allocated |
|---------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode          |    59.61 us |  1.155 us |  1.798 us | 0.1221 |   4.02 KB |
| CreateQRCodeMultiMode |   524.70 us |  2.772 us |  2.593 us |      - |   7.22 KB |
| CreateQRCodeLong      | 1,283.52 us | 25.074 us | 24.626 us |      - |  10.89 KB |
| CreateQRCodeLongest   | 7,074.98 us | 58.641 us | 48.968 us |      - |  43.83 KB |

## Sample Before - 664 bits encoded - version 5

```
██████████████  ██    ████    ██████  ██████████████        ██████████████
██          ██    ████          ██        ██  ██      ██    ██          ██
██  ██████  ██      ██  ██    ████  ██    ██  ██████        ██  ██████  ██
██  ██████  ██          ██      ██████  ██  ██  ██          ██  ██████  ██
██  ██████  ██    ██  ██    ████████      ████████  ██      ██  ██████  ██
██          ██      ██  ██    ██████    ██    ██  ████  ██  ██          ██
██████████████  ██  ██  ██  ██  ██  ██  ██  ██  ██  ██  ██  ██████████████
                ██  ████  ██  ████  ██  ██    ████████  ██                
████  ████  ██    ██████  ██      ████████      ██████████  ██          ██
████    ████    ████    ████    ██          ██  ██    ██████  ██  ████  ██
██        ██████      ██  ██  ████  ██  ██    ██  ██  ████  ██  ██    ████
  ██  ██████  ██  ██████████    ████    ██      ██  ██    ██████    ██  ██
██  ██      ██  ██    ██████  ████    ████  ████    ████                ██
  ████    ██    ██  ██    ████  ██  ████    ██  ██████    ████████      ██
██    ████  ██    ██████████    ██    ████  ████      ██    ████      ████
  ██      ██  ██████  ██      ██    ██████    ████████        ████    ████
  ██  ██  ████    ████      ██    ██        ████  ██  ████        ████  ██
        ████  ██  ██  ████████  ████        ██████  ████      ████  ██████
██████████  ██████  ████      ██████  ██    ██████      ██  ████  ██    ██
██  ██  ██    ████████  ██  ██      ██  ██████    ██████  ████████████████
██  ██      ██          ████████  ████████████        ████  ██████    ████
  ██  ████      ██  ██      ██  ██  ██  ██      ██  ████████    ██  ██  ██
██        ████  ████    ██  ████████          ██████  ████  ██  ████████  
██  ████  ██  ████          ████████████  ██  ██  ██  ██████  ██  ██  ████
████████  ██████████      ████████    ██  ██████  ██  ██  ██    ██      ██
████  ██  ██  ██    ██  ██    ██    ██████  ██  ████████    ██  ████    ██
████████    ████  ████      ████████████    ████    ████  ████    ████  ██
██  ██  ██    ████      ██████            ██  ████████████    ████████████
██          ██    ██  ████    ████████  ████  ██████  ████████████  ██  ██
                ██    ██      ██  ██████    ██████  ██████      ██████████
██████████████          ████████  ██  ████  ██        ████  ██  ██████  ██
██          ██        ██  ██  ██    ████    ██████████████      ██  ████  
██  ██████  ██  ██    ██  ██  ██  ████████  ██  ██    ████████████  ██  ██
██  ██████  ██  ██  ██  ████    ██  ██  ██    ██    ██      ██  ██████  ██
██  ██████  ██    ██  ██  ██  ██    ████  ████████████      ██        ████
██          ██  ██    ██████      ██  ██    ████  ██        ██    ████    
██████████████  ██    ██████  ████  ████  ██████          ██    ██████  ██
```

## Sample After - 526 bits encoded - version 4

```
██████████████  ██  ██  ██    ████████          ██  ██████████████
██          ██    ████████  ██████  ██  ██  ██      ██          ██
██  ██████  ██  ██████    ██  ██    ████  ████  ██  ██  ██████  ██
██  ██████  ██  ██████  ████    ██  ██  ██    ████  ██  ██████  ██
██  ██████  ██  ██    ██████████  ██    ████    ██  ██  ██████  ██
██          ██      ████  ████  ██  ████  ██  ████  ██          ██
██████████████  ██  ██  ██  ██  ██  ██  ██  ██  ██  ██████████████
                    ██  ██  ██  ████    ██    ████                
████████    ██  ██  ██████    ██  ████████    ██  ██    ██████  ██
████  ██            ██  ██  ████  ████  ████  ██    ██    ██  ████
  ██████  ████████████████████  ████        ████  ██    ██      ██
██  ████████    ██████    ██    ████████  ██            ██████████
██    ██  ██████  ████  ██    ██    ██████  ██████    ██  ██████  
  ██  ██████    ████  ████      ██  ██  ████  ████  ██  ████    ██
  ██  ██    ██  ██  ████  ████    ██████            ██      ██    
████            ████  ██      ██    ████████  ████        ██  ██  
  ██████████████  ████  ████    ██      ██  ██      ████  ████████
██      ██    ██  ██      ██      ██    ████████  ████████    ████
████    ██████  ██████        ████    ██  ████  ██████  ██    ██  
██  ████████    ████████    ██████████    ████    ████████████    
  ██  ████  ████  ████    ██████████████    ████    ██      ██████
████  ██  ██  ██████                ██  ██    ██        ██████  ██
    ██  ██  ██  ██  ██      ██      ████    ██  ████              
  ██████  ██  ████  ████████        ████████    ████  ██████      
██    ████  ██  ████████      ██  ████      ██  ██████████████    
                ██    ██  ████            ██  ████      ██    ██  
██████████████          ██  ████    ██  ██      ██  ██  ████  ██  
██          ██          ████  ██  ██████  ██  ████      ██████  ██
██  ██████  ██    ██████            ██      ████████████████      
██  ██████  ██  ██      ████  ██  ████  ██  ██  ██████  ████  ██  
██  ██████  ██  ████  ████            ████████████  ██    ██      
██          ██  ████    ██    ██        ██    ████████████████    
██████████████  ██  ████████    ██        ██████  ██████████████  
```

## Summary by CodeRabbit

* **New Features**
  * Smarter QR encoding that automatically switches between numeric, alphanumeric, and byte modes for mixed-content text.
* **Performance**
  * Faster QR generation for Latin‑1 (Western European) text under default settings via a fast-path optimization.
* **Tests**
  * Added approval tests validating mixed-mode encoding and ensuring mixed-mode does not expand encoded size unexpectedly.


Shane32:
@gfoidl  Should this be a minor version bump?  (to version 1.8.0 vs 1.7.1)  I assume so, just in case it has an unintended side-effect in an edge case, but what do you think?  It does not change the public API and meets QR code spec, and only is applied in scenarios where it is "known" to be compatible, so it could be simply a revision.

Shane32:
Note that the optimization algorithm is tailored for QR code versions 1-9.  For versions 10-26 or 27-40 there would ideally be different breakpoints at which the algorithm should switch.  This means for versions 10-40, it may be possible to write a carefully-constructed QR code which would take up more space than unoptimized.  The content would have to be tailored to trigger a mode change exactly at specific points while also maxing out the capacity for the requested version, such as this sample code, which throws a data too large exception:

```cs
var gen = new QRCodeGenerator();
var str = "111111a111111a111111a111111a111111a111111";
var s = new string('a', 119 - str.Length) + str;
_ = gen.CreateQrCode(s, ECCLevel.H, requestedVersion: 10);
```

The existing code is written to encode to a bit array before selecting the version.  And the selected version changes the size of the bit array, so it would be a problem to do it in the reverse order.  For a truly ideal algorithm, we would need to split the calculation of the ending length from the encoding to a bit array, so it could be calculated multiple times while determining the ideal version.

Shane32:
We could change the design so the most conservative breakpoints are used, such that it would not always be optimized for versions 1-26 but would never cause the issue described above in rare situations for versions 10-40.

Shane32:
@gfoidl I want to work on DataSegment some more before merging this - see #677

Shane32:
> Note that the optimization algorithm is tailored for QR code versions 1-9. For versions 10-26 or 27-40 there would ideally be different breakpoints at which the algorithm should switch. This means for versions 10-40, it may be possible to write a carefully-constructed QR code which would take up more space than unoptimized. The content would have to be tailored to trigger a mode change exactly at specific points while also maxing out the capacity for the requested version, such as this sample code, which throws a data too large exception:
> 
> ```cs
> var gen = new QRCodeGenerator();
> var str = "111111a111111a111111a111111a111111a111111";
> var s = new string('a', 119 - str.Length) + str;
> _ = gen.CreateQrCode(s, ECCLevel.H, requestedVersion: 10);
> ```
> 
> The existing code is written to encode to a bit array before selecting the version. And the selected version changes the size of the bit array, so it would be a problem to do it in the reverse order. For a truly ideal algorithm, we would need to split the calculation of the ending length from the encoding to a bit array, so it could be calculated multiple times while determining the ideal version.

Fixed; these notes are not applicable anymore.  The revised code adapts to the version being used, recalculating the bit length for each version breakpoint.

Shane32:
> We could change the design so the most conservative breakpoints are used, such that it would not always be optimized for versions 1-26 but would never cause the issue described above in rare situations for versions 10-40.

n/a

Shane32:
Increased memory usage is fixed

QRCoder/QRCodeGenerator.cs
```diff
@@ -112,6 +112,7 @@ public static QRCodeData GenerateQrCode(string plainText, ECCLevel eccLevel, boo
     {
         eccLevel = ValidateECCLevel(eccLevel);
         // Create data segment from plain text
+        // Use optimized segmentation when not forcing UTF-8, no BOM, and default ECI mode
```
Shane32:
```suggestion
```

QRCoder/QRCodeGenerator.cs
```diff
@@ -125,6 +125,10 @@ public static QRCodeData GenerateQrCode(string plainText, ECCLevel eccLevel, boo
     /// </summary>
     private static DataSegment CreateDataSegment(string plainText, bool forceUtf8, bool utf8BOM, EciMode eciMode)
     {
+        // Fast path: Use optimized Latin1 segment if conditions allow
+        if (!forceUtf8 && !utf8BOM && eciMode == EciMode.Default && OptimizedLatin1DataSegment.CanEncode(plainText))
+            return new OptimizedLatin1DataSegment(plainText);
+
```
Shane32, line 131: The new code should be effective in most usage scenarios, but does not change any semantics (whether correct or not) for strings that contain characters outside of ISO 8859-1, or when a specific ECI or UTF mode has been specified.

Shane32:
@gfoidl Ready for review.  On a side note, due to recent PRs, memory usage in the 'before' case has been reduced.

gfoidl (approved):
Now it reads really nice, and thanks for all the comments that make it easy to follow along.

> Should this be a minor version bump?

Yes. It brings in an enhancement, so minor should be increased here (depending on when you plan to release 1,8.0).

Or put in other words. That change shouldn't go into 1.7.x.

PR #681, opened 2025-10-14
Fix condition for ISO-8859-1 encoding selection in GetTargetEncoding method

### LINKED ISSUE #679: Fix text text encoding selection algorithm
When supplying ISO 8859-1 text to a QR code that has UTF-8 specified, the text will be encoded as ISO 8859-1 and still contain the UTF-8 marker.

Shane32:
## Summary by CodeRabbit

* **Bug Fixes**
  * QR code encoding now correctly respects the selected ECI mode.
  * ISO-8859-1 fallback is applied only when ECI mode is set to Default, preventing unintended overrides with non-default modes.
  * Ensures consistent behavior across UTF-8, BOM, and other encodings when a specific ECI mode is chosen.


gfoidl (approved):
Good bots 😄.

PR #684, opened 2025-10-14
Optimize alphanumeric encoder and galois field constants

Shane32:
## Summary by CodeRabbit

* **New Features**
  * Improved alphanumeric detection for more accurate encoding mode selection with mixed content.

* **Bug Fixes**
  * Corrected handling of certain non-numeric and special characters to ensure consistent encoding decisions.
  * Hardened IBAN validation to reduce false positives/negatives.

* **Refactor**
  * Optimized alphanumeric lookup and encoding tables for leaner, faster encoding.

* **Tests**
  * Expanded and simplified tests validating full alphanumeric mapping and encoding behavior.


QRCoder/QRCodeGenerator/AlphanumericEncoder.cs
```diff
@@ -7,38 +7,32 @@ public partial class QRCodeGenerator
...
-            // Add uppercase alphabetic characters.
-            for (char c = 'A'; c <= 'Z'; c++)
-                localAlphanumEncDict.Add(c, localAlphanumEncDict.Count);
-            // Add special characters from a predefined table.
-            for (int i = 0; i < _alphanumEncTable.Length; i++)
-                localAlphanumEncDict.Add(alphanumEncTable[i], localAlphanumEncDict.Count);
-            return localAlphanumEncDict;
-        }
+#if HAS_SPAN
+        private static ReadOnlySpan<byte> _map =>
```
Shane32: The `=>` is correct here.

QRCoder/QRCodeGenerator/AlphanumericEncoder.cs
```diff
@@ -7,38 +7,34 @@ public partial class QRCodeGenerator
...
-            for (char c = 'A'; c <= 'Z'; c++)
-                localAlphanumEncDict.Add(c, localAlphanumEncDict.Count);
-            // Add special characters from a predefined table.
-            for (int i = 0; i < _alphanumEncTable.Length; i++)
-                localAlphanumEncDict.Add(alphanumEncTable[i], localAlphanumEncDict.Count);
-            return localAlphanumEncDict;
-        }
+#if HAS_SPAN
+        // With C# 7.3 and later, this byte array is inlined into the assembly's read-only data section, improving performance and reducing memory usage.
+        // See: https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-core-3-0/
```
gfoidl, line 12:
```suggestion
        // See: https://vcsjones.dev/csharp-readonly-span-bytes-static
```
This link is more concise.

QRCoderTests/QRGeneratorTests.cs
```diff
@@ -379,14 +380,70 @@ public void can_encode_various_strings_various_ecc(int inputChars, ECCLevel eccL
     [Fact]
     public void validate_alphanumencdict()
     {
-        var gen = new QRCodeGenerator();
+        var encoderType = Type.GetType("QRCoder.QRCodeGenerator+AlphanumericEncoder, QRCoder");
```
gfoidl: I'd rather make the other types `private -> internal` and have an `InternalsVisibleTo` defined (in the csproj), so that reflection / expression tree stuff can be avoided. 

For test code the reflection thing is OK, because when something breaks the test will fail. But a more direct approach would be nicer.

Shane32: I agree

QRCoder/QRCodeGenerator/AlphanumericEncoder.cs
Shane32: True, but I'd prefer to link to official MS docs.

QRCoder/QRCodeGenerator/AlphanumericEncoder.cs
```diff
@@ -7,38 +7,34 @@ public partial class QRCodeGenerator
...
+            0, 1, 2, 3, 4, 5, 6, 7, 8, 9,
+            // 58..64  (: ; < = > ? @)
+            44, 255, 255, 255, 255, 255, 255,
+            // 65..90  (A..Z)
+            10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35
+            // (we don't index > 90)
+        ];
 
         /// <summary>
         /// Checks if a non-digit character is present in the alphanumeric encoding table.
```
Shane32:
```suggestion
        /// Checks if a character is present in the alphanumeric encoding table.
```

QRCoderTests/QRGeneratorTests.cs
Shane32: Changed reflection to use `internal` instead

QRCoder/QRCoder.csproj
```diff
@@ -16,6 +16,14 @@
     <Guid>e668b98b-83bb-4e60-b33c-4fd5ed9c0156</Guid>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(Configuration)' == 'Release'">
+    <InternalsVisibleTo Include="QRCoderTests, PublicKey=002400000480000094000000060200000024000052534131000400000100010071e74d3f6dcbba7726fcbb7ad05a2dfa3b8e13f9bf2938c05de266a83a7f3c95201b7cdc9e1f88842093ece868c3ec7874e3b8907008763c85711bf0e2e9757a3068385380a757bd52fa77248f227f602b0785e040756ef42dabc7de7f8376847c71b3f20356a9176cc88067583ee3501e61f8aa07bdd70f41418986fb79a1a3" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(Configuration)' != 'Release'">
+    <InternalsVisibleTo Include="QRCoderTests" />
+  </ItemGroup>
```
Shane32, line 25: What a pain!  A strongly-named assembly can only reference another strongly-named assembly.  If I make it conditional upon release, then the solution won't build in release mode because the tests can't see the internal members.  So I had to make the test project strongly named, and extract the public key to put in the prop here, and it had to be conditional because otherwise in debug mode it couldn't find the assembly with the strong name...

QRCoder/QRCodeGenerator/OptimizedDataSegment.cs
```diff
@@ -175,7 +175,7 @@ private static EncodingMode SelectInitialMode(string text, int startPos, int ver
                 if (numericCount < threshold)
                 {
                     var nextPos = startPos + numericCount;
-                    if (nextPos < text.Length && !IsAlphanumericNonDigit(text[nextPos]))
+                    if (nextPos < text.Length && !IsAlphanumeric(text[nextPos]))
```
Shane32, line 178: The next character is known to not be a digit, so it doesn't matter if this checks against IsAlphanumeric or IsAlphanumericNonDigit.  Since the new optimzation is faster with the former, it's changed here and below.

QRCoder/QRCoder.csproj
gfoidl: >  A strongly-named assembly can only reference another strongly-named assembly

Yeah, that virality sucks.

> be conditional because otherwise in debug mode

When it's strong named in DEBUG and RELEASE it should work in all cases?! So no conditional is needed.

But it's fine as is too.

Directory.Build.targets
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <!-- Strong name signing for Release builds of packable projects -->
-  <PropertyGroup Condition="'$(Configuration)' == 'Release' AND '$(IsPackable)' == 'true'">
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
```
gfoidl, line 10: This is what I referred to. Just make it unconditional signed.
But we won't change that often (never), so fine as is too.

gfoidl (approved):

QRCoder/QRCoder.csproj
Shane32: Codecov can’t inject its code into a strongly named assembly. So to get the coverage reports working I had to disable strong name signing.

Directory.Build.targets
Shane32: Codecov - See note above

QRCoder/QRCoder.csproj
gfoidl: Ah, I didn't think about this. Thanks for clarification.

PR #685, opened 2025-10-18
Prep for v2; target .NET Standard 2.0, .NET Standard 2.1 and .NET 6.0

### LINKED ISSUE #553: [WIP/QRCoder2] Which frameworks and language version to target
*__Note__: This issue is part of the planning of version 2 of the QRCoder. The [meta-issue can be found here](https://github.com/codebude/QRCoder/issues/512). These comments have been copied into this issue here, with the comments marked as such. If comments on the topic of this issue already exist in the meta thread, they have been copied here, naming the authors.*

### Topic
The current version of the QRcoder (1.X) supports a whole range of target frameworks. Among them also quite old versions like .NET3.5. Which frameworks do we want to target in QRCor version 2 and why (pro/contra)?

Shane32:
This PR:
- Targets the `develop` branch (for version 2.0)
- Adds a dependency on System.Memory for .NET Standard 2.0 projects (so `Span` and `ArrayPool` can be used)
- Eliminate dependency on System.Text.Encoding.CodePages (users will need to reference it manually when needed)
- Reduces as many `#if` blocks as possible, adding polyfills where necessary.
- Updates workflows and version number appropriately
- Removes the obsolete and unused `StringValueAttribute`
- Eliminate testing on .NET Core 2.1 and .NET 5.0 -- remaining tests on .NET Framework 4.6.2, .NET Core 3.1 and .NET Core 6.0 will test all three target tfms of the main project

Future PR:
- Move components that depend on `System.Drawing.Common` to another package

Then, the compilation that targets .NET Standard 2.1 and .NET 6.0 will be dependency-free.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed `StringValueAttribute` and `CustomExtensions` classes from public API
  * Dropped support for older .NET frameworks; minimum targets now .NET Standard 2.0, .NET Standard 2.1, and .NET 6.0

* **Chores**
  * Version updated to 2.0.0-preview
  * CI workflows updated; removed testing on obsolete framework versions


QRCoder/PdfByteQRCode.cs
```diff
@@ -55,120 +55,105 @@ public byte[] GetGraphic(int pixelsPerModule, string darkColorHtmlHex, string li
...
         using var stream = new MemoryStream();
-#if NETFRAMEWORK
-        var writer = new StreamWriter(stream, System.Text.Encoding.ASCII);
-#elif NET6_0_OR_GREATER
-        var writer = new StreamWriter(stream, System.Text.Encoding.ASCII, leaveOpen: true);
-#else
-        var writer = new StreamWriter(stream, System.Text.Encoding.ASCII, 1024, true);
-#endif
-        try
+        using var writer = new StreamWriter(stream, Encoding.ASCII, 1024, true);
```
Shane32, line 58: Added `using` and removed `try..finally` which changed indentation.

Shane32:
@coderabbitai review

Shane32:
Note: I still plan to issue another v1 release from the `master` branch.

QRCoder/PdfByteQRCode.cs
```diff
@@ -55,120 +55,105 @@ public byte[] GetGraphic(int pixelsPerModule, string darkColorHtmlHex, string li
...
+            "/Kids [ <<\r\n" +                                                      // Array of page objects - begin inline page dictionary
+                "/Type /Page\r\n" +                                                 // Declares this as a page
+                "/Parent 2 0 R\r\n" +                                               // References parent Pages object
+                "/MediaBox [0 0 " + pdfMediaSize + " " + pdfMediaSize + "]\r\n" +   // Page dimensions [x1 y1 x2 y2]
+                "/Resources << /ProcSet [ /PDF ] >>\r\n" +                          // Required resources: PDF operations only (no images)
+                "/Contents 3 0 R\r\n" +                                             // References content stream (object 3)
+                ">> ]\r\n" +                                                        // End inline page dictionary and Kids array
+            ">>\r\n" +                                                              // End dictionary
+            "endobj\r\n"                                                            // End object
+        );
```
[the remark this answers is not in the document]
Shane32: #687

QRCoder/PdfByteQRCode.cs
gfoidl: This could be a static ROS<byte> property, so that no array needs to be created as Roslyn will refer to the assembly's static data segment.

As System.Memory package is there, I assume Roslyn will also emit the optimized code (i.e. no allocation) for the older targets.

When not possible, this should be at least a `static readonly` field.

QRCoder/PdfByteQRCode.cs
```diff
@@ -55,120 +55,105 @@ public byte[] GetGraphic(int pixelsPerModule, string darkColorHtmlHex, string li
...
+
+        // PDF header - declares PDF version 1.5
+        writer.Write("%PDF-1.5\r\n");
+        writer.Flush();
+
+        // Binary comment - ensures PDF is treated as binary file (prevents text mode corruption)
+        stream.Write(_pdfBinaryComment, 0, _pdfBinaryComment.Length);
+        writer.WriteLine();
+
+        writer.Flush();
```
gfoidl, line 69: Is flushing needed? We write to a memory stream, so I don't see real benefit from it.

Same on the other flushes below.

QRCoder/PdfByteQRCode.cs
```diff
@@ -55,120 +55,105 @@ public byte[] GetGraphic(int pixelsPerModule, string darkColorHtmlHex, string li
...
+            "<<\r\n" +                                         // Begin trailer dictionary
+            "/Size " + ToStr(xrefs.Count + 1) + "\r\n" +       // Total number of entries in xref table
+            "/Root 1 0 R\r\n" +                                // Reference to catalog object
+            ">>\r\n" +                                         // End trailer dictionary
+            "startxref\r\n" +                                  // Start of xref keyword
+            ToStr(startxref) + "\r\n" +                        // Byte offset of xref table
+            "%%EOF"                                            // End of file marker
+        );
+
+        writer.Flush();
```
gfoidl, line 155: This flush is fine, as w/o disposing of the StreamWriter not all data written may end up in the (Memory) stream.

gfoidl (commented):
Nice cleanup!

QRCoder/PdfByteQRCode.cs
Shane32: Done

QRCoder/PdfByteQRCode.cs
Shane32: It's the text writer that requires flushing, not the memory stream.  Without flushing, the position may be inaccurate.  Now since we are using ASCII encoding, there is (probably) no reason for the encoder to cache characters, but I don't know that it says anywhere that the ASCII encoder doesn't require flushing.  So the encoder flushes before reading the position property.

QRCoder/PdfByteQRCode.cs
Shane32: If we disposed of the stream writer, then we could omit the flush, as flushing is implicit during dispose.  However, we are using the shorter using syntax so the flush is required here to ensure all characters have been written.

gfoidl: Here we are talking about the same. Maybe I was unclear.

gfoidl (approved):

PR #698, opened 2026-07-12
Minor performance improvement in BitmapByteQRCode + clarify padding calculation

### LINKED ISSUE #697: Bug in BMP padding calculation when pixelsPerModule is odd
`BitmapByteQRCode.GetGraphic` uses the following formula to calculate the number of padding bytes per row:

```cs
var paddingLen = sideLength % 4;
```

That's not correct. It happens to work when `sideLength` is even, i.e. when `pixelsPerModule` is even, but it fails when `sideLength` is odd.

The correct formula would be `(4 - (sideLength % 4)) % 4`, or the more efficient equivalent `-sideLength & 3`.

KrisVandermotten:
## Summary

* [x] Fixes a bug in BMP padding calculation when `pixelsPerModule` is odd.
* [x] Minor optimization using [BinaryPrimitives.WriteInt32LittleEndian(Span<Byte>, Int32)](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives.writeint32littleendian) on modern .NET

## Closing issues
Fixes #697 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR bitmap generation across supported platforms for more consistent output.
  * Corrected bitmap row padding calculations to ensure proper byte alignment.
  * Enhanced binary integer writing on newer .NET Core versions to reliably produce the expected byte order.

### DIFF
--- a/QRCoder/BitmapByteQRCode.cs
+++ b/QRCoder/BitmapByteQRCode.cs
@@ -1,3 +1,8 @@
+#if NETCOREAPP2_1_OR_GREATER
+using System.Buffers.Binary;
+#endif
+using System.Diagnostics;
+
 using static QRCoder.QRCodeGenerator;
 
 namespace QRCoder;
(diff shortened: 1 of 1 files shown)


Shane32:
> * Fixes a bug in BMP padding calculation when `pixelsPerModule` is odd.

Note: An extensive analysis of the old and new calculation code by ChatGPT determined that it is mathematically equivalent for 24bpp images, but the new calculation better represents the intent, and is more correct if the data wasn't 24bpp.

Shane32 (approved):

PR #699, opened 2026-08-08
Optimize AlignmentPatterns

KrisVandermotten:
## Summary

This PR applies a round of optimizations to the calculation and placement of alignment patterns.

First, several improvements reduce the amount of memory allocated by `CreateAlignmentPatternTable`:

- The `AlignmentPattern` structure has been removed. It was wrapper around a list of Points and a `Version` property. As the `Version` property wasn't used, there is no need for the structure.
- The `Point` structure has been reduced from two `int` properties to two `byte` properties. Given that the result of the method is essentially a table of Points, this reduces the amount of memory needed to store the actual data by a factor 4.
- The resulting table has been changed from a Dictionary to an array.
- The entries in the table have been changed from `List<Point>` to `Point[]`.

All of the above contribute to less memory being needed to store the result and keep it referenced for later usage. In addition, some memory allocations have been reduced in the process to produce that result:

- On modern .NET, the allocation (and initialization) of `alignmentPatternBaseValues` is avoided. It's values are now bytes instead of ints.
- A single `List<Point>` is used as a buffer to store intermediate results, instead of a list per version.

Some changes have been made to reduce execution time:

- The calculation for version 1 is avoided, as it's empty anyway.
- The inner loops are terminated as soon as they can, instead of always looping 7 times.
- The unnecessary check whether a calculated point was not already present in the `points` intermediate list has been removed. As a result, the `IEquatable<Point>` implementation is no longer needed and removed.
- The outer loop is now a simple loop over the versions. Not only does this improve the clarity of the code, it also avoids an integer division per version.
- The numbers in `alignmentPatternBaseValues` are now the coordinates of the upper left corners of the alignment patters, not the center points. This saves two substractions per point.

I ran a simple benchmark to measure the results.

```
BenchmarkDotNet v0.13.12, Windows 11 (10.0.26200.7462)
11th Gen Intel Core i7-1165G7 2.80GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK 10.0.302
  [Host]     : .NET 8.0.29 (8.0.2926.32403), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 8.0.29 (8.0.2926.32403), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
```

On master:

| Method                      | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|---------------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| CreateAlignmentPatternTable | 12.63 us | 0.225 us | 0.211 us | 3.3722 | 0.2441 |  20.74 KB |

This PR:

| Method                      | Mean     | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|---------------------------- |---------:|----------:|----------:|-------:|-------:|----------:|
| CreateAlignmentPatternTable | 1.811 us | 0.0242 us | 0.0226 us | 0.5455 | 0.0057 |   3.35 KB |

Memory allocations have been reduced by a factor 6. Even though the method is executed only once, it's nice to see that it is now 7 times faster. More optimizations are possible but deemed unnecessary, as they would also reduce the readability of the code.

Secondly, some improvements have been made to the `PlaceAlignmentPatterns` method.

Obviously, placement already benefits from the fact that `AlignmentPatterns.FromVersion` now returns a `Point[]` instead of (a struct around) a `List<Point>`, and does so by indexing into an array instead of a dictionary.

On top of that:

- The `y` and `x` loops have been swapped to improve data locality, and avoid redundant indexing into the `ModuleMatrix` array.
- The inner `x` loop has been unrolled, primarily to reduce branches and branch mispredictions.
- A redundant construction of `alignmentPatternRect` is avoided.

The results are visible in the existing `QRCodeGeneratorBenchmark`.

On master:

| Method                | Mean        | Error     | StdDev    | Gen0   | Allocated |
|---------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode          |    126.8 us |   2.52 us |   3.28 us | 0.6104 |   4.02 KB |
| CreateQRCodeMultiMode |    984.3 us |  19.62 us |  31.69 us |      - |   7.22 KB |
| CreateQRCodeLong      |  2,330.6 us |  43.59 us |  70.39 us |      - |  10.89 KB |
| CreateQRCodeLongest   | 13,594.9 us | 268.84 us | 298.82 us |      - |  43.84 KB |

This PR:

| Method                | Mean        | Error     | StdDev    | Gen0   | Allocated |
|---------------------- |------------:|----------:|----------:|-------:|----------:|
| CreateQRCode          |    121.9 us |   2.43 us |   4.96 us | 0.6104 |   4.02 KB |
| CreateQRCodeMultiMode |    924.9 us |  18.14 us |  20.90 us | 0.9766 |   7.22 KB |
| CreateQRCodeLong      |  2,141.0 us |  41.43 us |  46.04 us |      - |  10.89 KB |
| CreateQRCodeLongest   | 12,644.3 us | 193.28 us | 171.34 us |      - |  43.84 KB |

## Test plan

All modified code is exercised by existing unit tests. Benchmarks show the performance benefit.

## Improvements

* Optimized QR code generation for faster, more efficient alignment pattern processing.
* Reduced memory usage when handling alignment coordinates.
* Improved alignment pattern generation across supported QR code versions, including Micro QR variants.
* Preserved the existing 5×5 alignment pattern appearance and placement.
* Added validation to help ensure alignment data remains within expected limits.
* Streamlined error-correction processing without changing generated QR code behavior.


QRCoder/QRCodeGenerator/AlignmentPatterns.cs
```diff
@@ -10,57 +12,64 @@ private static class AlignmentPatterns
...
-            var localAlignmentPatternTable = new Dictionary<int, AlignmentPattern>(40 + 4);
+            var localAlignmentPatternTable = new Point[4 + 1 + 40][];
+
+            // Micro QR codes do not have alignment patterns.
+            Point[] empty = [];
+
+            localAlignmentPatternTable[-4 + 4] = empty;
+            localAlignmentPatternTable[-3 + 4] = empty;
+            localAlignmentPatternTable[-2 + 4] = empty;
+            localAlignmentPatternTable[-1 + 4] = empty;
```
gfoidl: Can this indexing be simplified?

gfoidl (commented):
Quoting [coderabbitai](https://github.com/apps/coderabbitai):

> No actionable comments were generated in the recent review. 🎉

:smiley: one AI hasn't found something reviewing a other AI[^1]...(same model in the background?).

[^1]: note the "Summary by CodeRabbit" in the TOP got added later on by CodeRabbit.

KrisVandermotten:
> 😃 one AI hasn't found something reviewing a other AI...(same model in the background?).

Not sure what "other AI" you're referring to? No AI was used in making this PR, nor its description.

QRCoder/QRCodeGenerator/AlignmentPatterns.cs
KrisVandermotten: What simplification do you have in mind?

Note that the additions are compiled away by the C# compiler (constant folding).

gfoidl: > (constant folding)

Sure, but indices like `-4 + 4` are strange to read and non-obvious, where it just could be `[0]`, `[1]`, ...

KrisVandermotten: The `+ 4` matches the `+ 4` in lines 23, 44 and 69. Would it be more clear if I declared a `const int indexOffset = 4;` and used it throughout?

KrisVandermotten: Pushed a commit doing exactly that.

Shane32:
> * The unnecessary check whether a calculated point was not already present in the `points` intermediate list has been removed. As a result, the `IEquatable<Point>` implementation is no longer needed and removed.

Makes sense, but should we leave the IEquatable implementation anyway?  It really hurts performance if needed and missing.  See https://github.com/Shane32/QRCoder/pull/509#issuecomment-2092077502  But if unneeded we can remove it ....

QRCoder/QRCodeGenerator/AlignmentPatterns.cs
Shane32: Looks good

KrisVandermotten:
> > * The unnecessary check whether a calculated point was not already present in the `points` intermediate list has been removed. As a result, the `IEquatable<Point>` implementation is no longer needed and removed.
> 
> Makes sense, but should we leave the IEquatable implementation anyway? It really hurts performance if needed and missing. See [#509 (comment)](https://github.com/Shane32/QRCoder/pull/509#issuecomment-2092077502) But if unneeded we can remove it ....

The `QRCodeGenerator.Point` struct is `private`, so there is complete control over where and how it is used. It's `Equals` and `GetHashCode` methods aren't called. There are other such structs, e.g. `QRCodeGenerator.ECCInfo` and `QRCodeGenerator.CodewordBlock` that also don't have an `IEquatble<T>` implementation. That's why I removed the unused methods in `QRCodeGenerator.Point` as well.

Finally, benchmarks show an improvement in performance, not a regression.

Shane32:
Fair.  Just indicating that `Point.Equals` wasn't previously called directly; it was a hidden performance issue due to the call to `List<>.Contains`.  Let's at least add a comment to the struct to that effect, something like:

> Be sure to add an `IEquatable<>` implementation if calling `List<Point>.Contains` or similar methods requiring equality comparison.

to help prevent the same mistake in the future.

KrisVandermotten:
> Fair. Just indicating that `Point.Equals` wasn't previously called directly; it was a hidden performance issue due to the call to `List<>.Contains`.

I do understand. That's why I wrote: "The unnecessary check whether a calculated point was not already present in the points intermediate list has been removed. _As a result_, the IEquatable<Point> implementation is no longer needed and removed."

> Let's at least add a comment to the struct to that effect, something like:
> 
> > Be sure to add an `IEquatable<>` implementation if calling `List<Point>.Contains` or similar methods requiring equality comparison.
> 
> to help prevent the same mistake in the future.

Just to be clear: there are nine such structs, do you suggest to add such a comment to all of them, or just to the `Point` struct?

Shane32:
Let's just add a comment for the Point struct and call it a day.

Shane32:
Thanks! I'll merge once tests pass.  Thanks for your help!!

Shane32:
Sorry I forgot to merge this.  Can you update it so I can?

KrisVandermotten:
> Sorry I forgot to merge this. Can you update it so I can?

Done

PR #702, opened 2026-08-23
Cleanup and optimize CalculateGeneratorPolynom

KrisVandermotten:
## Summary

I set out to simplify the calculation of the generator polynomial for the error correction codes. Especially the polynomial multiplication was overly complex and inefficient. So I wrote a new clean implementation, running it in parallel with the existing one to make sure I got the same results. But then I realized that we don't need runtime calculation of these polynomials at all: they're always the same!

So I generated 30 polynomials, and embedded their coefficients in the assembly. I turned a simple version of the rewritten generation code into a unit test, thereby also documenting the generator code. I did have to make the `Polynom` and `PolynomItem` structs internal to make that possible, but I guess that's OK.

To be honest, only 18 polynomials will ever be used, but it turned out to be simpler (also for runtime performance) to include all 30, than to write code to select the correct coefficients for each of the 18 ones we need.

Other than replacing a few redundant range checks by `Debug.Assert` statements, I did not change the `Polynom` and `PolynomItem` data structure (yet).

The result is a significant reduction in the amount of code. More than 200 lines of complex code could be removed. On master, a .NET 6 release build is 201 KiB. On the PR branch, a .NET 6 release build is 198.5 KiB. That's a reduction of 2.5 KiB, despite the addition of the 0.5 KiB coefficient table.

## Test plan

All code is covered by existing tests. During development, I ran both old an new code, comparing results with `Debug.Assert` checks.

In addition, a test has been added to validate the correctness of the generator polynomial coefficients.

Update: despite what coderabbit.ai claims below, no bugs were fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR code error-correction calculations for more reliable generation.
  * Added safeguards for invalid error-correction settings and polynomial values.

* **Quality Improvements**
  * Improved consistency across QR code error-correction levels.
  * Strengthened validation during QR code generation.
  * Expanded automated checks for error-correction and Galois field calculations.


QRCoder/QRCodeGenerator/GaloisField.cs
```diff
@@ -16,47 +14,31 @@ public partial class QRCodeGenerator
...
-        /// Normalizes a Galois field exponent to ensure it remains within the bounds of the field's size.
-        /// This is particularly necessary when performing multiplications in the field which can result in exponents exceeding the field's maximum.
-        /// </summary>
-        public static int ShrinkAlphaExp(int alphaExp)
-        {
-            Debug.Assert(alphaExp >= 0);
-            return (int)((uint)alphaExp % 256 + (uint)alphaExp / 256);
-        }
+        public static byte GetAlphaExpFromIntVal(byte intVal)
+            => _alphaExponentByIntegerValue[intVal];
```
Shane32, line 42: I'd like to `Debug.Assert(intVal != 0)` with the comment `// Zero is not valid as it does not have an exponent representation.`

Shane32:
So if I understand this correctly, `CalculateGeneratorPolynom(int numEccWords)` always returns one of a small set of polynominals (since `numEccWords` is a small set) which can be precalculated and embedded, which is what eliminates most of the code in this PR.  Correct?

QRCoder/QRCodeGenerator/GaloisField.cs
KrisVandermotten: The problem with that is line 699 in QRCodeGenerator.cs

It used to be
```cs
index0Coefficient = index0Coefficient == 0 ? 0 : GaloisField.GetAlphaExpFromIntVal(index0Coefficient);
```

Thanks to the removal of the check, it can be simplified to 
```cs
index0Coefficient = GaloisField.GetAlphaExpFromIntVal((byte)index0Coefficient);
````

KrisVandermotten:
> So if I understand this correctly, `CalculateGeneratorPolynom(int numEccWords)` always returns one of a small set of polynominals (since `numEccWords` is a small set) which can be precalculated and embedded, which is what eliminates most of the code in this PR. Correct?

Correct.

On top of that, the now eliminated code was very complex and inefficient. Compare it to the equivalent that now lives in a unit test.

Shane32:
This now has eliminated a substantial amount of calculation code.  It seems that if we merge this PR, then if we want to support additional QR Code variants (see https://www.qrcode.com/en/codes/ ) then we may need to manually recalculate these values if the input changes.  Would it be wise to move this into the tests, such that the test calculates and verifies the precalculated values are correct?  Or just move it into a console app to generate the values or something?

KrisVandermotten:
> Would it be wise to move this into the tests, such that the test calculates and verifies the precalculated values are correct?

That's exactly what I did.

Shane32:
In other words, it seems that it would be a good idea to retain the code that was used to generate the values that are now constants embedded into the library.

Shane32:
> > Would it be wise to move this into the tests, such that the test calculates and verifies the precalculated values are correct?
> 
> That's exactly what I did.

Ok let me review further.

Shane32:
I see; so it's just greatly simplified in the tests

QRCoder/QRCodeGenerator/GaloisField.cs
Shane32: Got it.

Shane32 (approved):

Shane32:
This is great and makes complete sense - thanks !!!!

